PR #3, opened 2019-03-24
Update uk.m3u

desiflex:
add


desiflex:
add

raieton:
Could you also update french channels? It will nice to have at least all free channels :wink:

PR #22, opened 2019-05-09
Test only a specific country code via TEST_COUNTRY=xx env variable

mephist:
Hello,

It must be useful to check only a specific playlist which someone is working on.


freearhey:
Great idea. I will check the code and publish it to master branch in the next update.

PR #23, opened 2019-05-10
Add iOS Information

tiagomartinho:
I verified that the information shared is compatible with the app I'm developing. 

Let me know if it's helpful!


tiagomartinho:
Thank you!

PR #239, opened 2019-08-21
Update and rename channels/ae.m3u to  channels/ae.m3u

luseildes:


freearhey:
I don't get it, you just replaced the EPG url with a colon. What for?

PR #323, opened 2019-09-08
Update +CHANNEL REQUESTS+.txt

boomski:
Hi, could we have a list like this with Channel requests.
It woulld be great to have an overview on  channels to search for.
If we remove deadlinks we should add them to this list ,so we can later add them back when there is a new link for these channels and have an overview of missing channels.
Made an example , i think it should make the work bit easier.

Greetings, boomski


freearhey:
@boomski Hello! I wanted to offer a slightly different solution. What about using the GitHub Project? There is a search and the ability to track all requests. Here is an example: https://github.com/freearhey/iptv/projects/1

boomski:
Hi, Yes this is good solution
The GitHub Project is perfect for this 
Thanks!

freearhey:
@boomski Just one thing, make sure you can add and edit lists in the project. I thought I gave you permission, but maybe something else needs to be turned on.

PR #332, opened 2019-09-08
Update at.m3u

boomski:
Added Servus TV from channels request in 3 qualities
This stream will only work in Austria or with a Austrian VPN connection. (Geoblock)


freearhey:
@boomski I think we should not include geoblocked broadcasts in playlists. It’s easier to remove it from the list than to support it in the future.

And yes, feel free to move the cards in Project to the "Done" list as soon as you send the PR, just so you don't have to write about it here.

boomski:
I would like to help you with the cards , but i can't move or add them , the only thing i can do in the projects page is copy link of the card , I think you need to give me access as a collaborator to help you with that.

freearhey:
@boomski Oh. I thought I sent you an invite today, but apparently something did not work. Here, try this link: https://github.com/freearhey/iptv/invitations And if this also doesn’t work, then let me know and I will try something else.

boomski:
@freearhey Yes , now it works

freearhey:
@boomski Great!

PR #336, opened 2019-09-08
Update al.m3u

boomski:
Updated link  Appolon TV


freearhey:
@boomski Please make all changes only in the dev branch. The thing is, before updating the master branch, I need to generate grouped playlists (index.full.m3u, index.sport.m3u, etc.) from the dev branch and this means that changes from master may not get into these playlists.

PR #394, opened 2019-10-03
Update ec.m3u

boomski:
Added 4 Channels


channels/ec.m3u
```diff
@@ -1,15 +1,21 @@
 #EXTM3U
-#EXTINF:-1 tvg-id="" tvg-name="" tvg-logo="https://i.imgur.com/ATPWac1.jpg" group-title="",Ecuador TV
+#EXTINF:-1 tvg-id="" tvg-name="" tvg-logo="https://i.imgur.com/zDhynPe.png" group-title="",Ecuavisa Canal 2 
+https://youtvplay.com/static/wowza/test/hls/20001/360p/playlist.m3u8
```
freearhey: This is not a live stream. Shows the same thing every time it starts.

freearhey (changes requested):

channels/ec.m3u
freearhey: @boomski You can remove these two lines directly from the pull request and I could immediately merge it.

boomski: @freearhey  I Deleted the 2 lines

freearhey (approved):

PR #437, opened 2019-10-06
Update ie.m3u

boomski:


channels/ie.m3u
```diff
@@ -1,7 +1,21 @@
...
+#EXTINF:-1 tvg-id="" tvg-name="" tvg-logo="https://i.imgur.com/DRPAvjX.jpg" group-title="",Seanad Éireann
+https://d1h8rkqumz239n.cloudfront.net/seanad/ngrp:seanad.stream_all/chunklist.m3u8
+#EXTINF:-1 tvg-id="" tvg-name="" tvg-logo="https://i.imgur.com/RNVkCaY.jpg" group-title="",Oireachtas Committee Room 1
+https://d1h8rkqumz239n.cloudfront.net/cr1/ngrp:cr1.stream_all/chunklist.m3u8
+#EXTINF:-1 tvg-id="" tvg-name="" tvg-logo="https://i.imgur.com/ZQt5jvM.jpg" group-title="",Oireachtas Committee Room 2
+https://d1h8rkqumz239n.cloudfront.net/cr2/ngrp:cr2.stream_all/chunklist.m3u8
+#EXTINF:-1 tvg-id="" tvg-name="" tvg-logo="https://i.imgur.com/5d6WOPx.jpg" group-title="",Oireachtas Committee Room 3
+https://d1h8rkqumz239n.cloudfront.net/cr3/ngrp:cr3.stream_all/chunklist.m3u8
+#EXTINF:-1 tvg-id="" tvg-name="" tvg-logo="https://i.imgur.com/odMa1KI.jpg" group-title="",Oireachtas Committee Room 4
+https://d1h8rkqumz239n.cloudfront.net/cr4/ngrp:cr4.stream_all/chunklist.m3u8
```
freearhey: Something seems wrong with these streams. They seem to start, but without an image or sound and after 9 seconds they stop working at all.

freearhey (changes requested):

channels/ie.m3u
boomski: - I tested the links on streamtest.in -> OK
- Tested the streams in Kodi -> OK
- Tested them in VLC player and got the problem you mentioned. , so they seem not to play when you use VLC player.

freearhey: Let’s then agree that we will consider the broadcast to be working only if it is simultaneously launched in both the VLC player and Kodi. As I understand it, these are the most popular free players with support IPTV. Is that all right?

boomski: It's all right for me

freearhey (approved):

PR #450, opened 2019-10-08
Sorted channels

freearhey:
Sorted channels in the international playlist by country.


channels/se.m3u
```diff
@@ -23,3 +23,5 @@ https://csm-e-tv4se-eb.tls1.yospace.com/csm/live/142791553.m3u8
 https://edg01-prd-se-dcs.solidtango.com/edge/_lo9yf4l5_/lo9yf4l5/playlist.m3u8
 #EXTINF:-1 tvg-id="" tvg-name="" tvg-logo="https://i.imgur.com/O1wCZTZ.jpg" group-title="Local",Öppna Kanalen
 http://83.140.64.214/edge/_451iw2h_/451iw2h/playlist.m3u8
+#EXTINF:-1 tvg-id="" tvg-name="" tvg-logo="https://i.imgur.com/QpdHuGo.png" group-title="Entertainment",MTV Sweden
+http://unilivemtveu-lh.akamaihd.net/i/mtvno_1@346424/index_3500_av-b.m3u8
```
boomski: should be MTV Norge and placed in Norwegian playlist.

boomski (changes requested):

channels/se.m3u
boomski: https://no.wikipedia.org/wiki/MTV_Norge

freearhey: and I thought so too) but then it turned out that the MTV Norge channel was closed in 2005 and was replaced by MTV Sweden (https://en.wikipedia.org/wiki/MTV_(Swedish_TV_channel))

boomski: MTV Norge was not closed in 2005  , It was MTV Nordic who was closed which is not the same.
In 2005 MTV Nordic Closed and they Launched 4 New Channels , 1 for every Nordic Country : Sweden , Norway , Finland and Denmark. 

Every of these channels are subtitled in their own Language , and the link we got here the subs are in Norwegian Language so my gess is this is MTV Norway (Also known as MTV Norge).

freearhey: Yes you are right! My mistake.

I’ve already moved the channel to no.m3u

boomski (approved):

PR #478, opened 2019-10-10
Update co.m3u

boomski:


madtom75:
Hi I’m desperately looking for KTLA5 and KCAL9 IPTV, any help?

channels/co.m3u
```diff
@@ -65,3 +65,5 @@ http://api.new.livestream.com/accounts/28238881/events/8574038/live.m3u8
 https://live-edge-rbx-1.cdn.enetres.net/184784E1D210401F8041E3E1266822CC021/live-1500/index.m3u8
 #EXTINF:-1 tvg-id="" tvg-name="" tvg-logo="https://i.imgur.com/kady79t.jpg" group-title="",UAO Play +
 https://live-edge-sxb-1.cdn.enetres.net/184784E1D210401F8041E3E1266822CC023/live-1500/index.m3u8
+#EXTINF:-1 tvg-id="" tvg-name="" tvg-logo="https://i.imgur.com/SknQ25w.png" group-title="",RCN [KODI]
+https://rcnhd.cdn.net.co/live01/rcnhd_720p.m3u8
```
freearhey: does not work in VLC player

freearhey (changes requested):

boomski:
Oke, but nothing is wrong with the stream because on other platforms it plays , it is just on VLC it doesn't work. so we could add it  with a tag behind the name [NO VLC] , like we did in hr.m3u playlist where a lot of channels don't play in Kodi they are tagged  [NO KODI] , and they work in VLC .
What do you think ?

freearhey:
@boomski I thought we decided not to add translations that do not work in VLC or Kodi. 
As for the channels in hr.m3u, in this case I think they should be deleted. Unless you mind of course.

boomski:
I should keep the channels in the lists in hr.m3u they work and it says NO KODI so users know it will not work on the mentioned platform. It is a little bit the same discussion as the geoblocked url maybe we should ask other users their opinion about this and which app or program they use for watching?

freearhey:
Ok, let's ask others.

PR #626, opened 2019-11-03
Channel Languages Updated for UAE

themak1985:
Many were listed as english - actually arabic.


freearhey:
@themak1985 Many of the languages were put down automatically based on the attached EPG or on information from the wikipedia, which is why there are such obvious errors. And that's why your help is twice as useful!

freearhey (approved):

PR #662, opened 2019-11-09
Update ir.m3u

xmha97:
Updated EPG source


freearhey (changes requested):
Please correct the playlist according to these instructions: https://github.com/iptv-org/iptv/blob/master/.github/CONTRIBUTING.md#add-epg-source so that we can approve the PR.

freearhey:
I checked the program more carefully and as I see it doesn't actually have a current program for any of the playlist channels. So I have to reject this PR. But anyway, thank you for your contribution!

PR #696, opened 2019-11-17
Added missing Estonian tvg-language

themak1985:
Could not get stream to work - but based on what I can tell - its estonian


freearhey:
It's actually a Russian-language channel. But anyway, the broadcast doesn't work, so we can just delete it.

PR #704, opened 2019-11-17
Updated Iraqi tvg-language, streams and TVG ID

themak1985:
Note - some of these channels broadcast in Syriac (in addition to arabic) - however - Syriac does not have an iso-639-1 language code. It does have a iso 639-2 code. In the meantime - updated to Arabic.


freearhey (approved):

freearhey:
Yes, I've faced this situation before, too. A way out of this situation would be to add support for ISO-639-2 to the code, but unfortunately there is no time for that. Maybe sometime in the future.

PR #782, opened 2019-12-08
Update LB logos and EPG data

themak1985:


freearhey (changes requested):
Could you provide a link to the EPG from which you took the data?

freearhey (approved):
Okay, I found the source of the EPG myself. But next time please provide a link to the EPG from which you take the data (here's the guide: https://github.com/iptv-org/iptv/blob/master/.github/CONTRIBUTING.md#add-epg-source), otherwise no one will be able to use it.

themak1985:
Sorry about that, will do.

PR #851, opened 2019-12-24
Update Some Channel Links

MapGuy11:
The Country Network: Recently Went To A Pay Wall I Have A Antenna Signal For There Channel even though its not M3U8 url it is just disguised
Added WPHL a MY TV network affiliate
Added WPHL.2 Antenna TV affiliate

I have more links Just look at my IPTV list https://pastebin.com/raw/EhANLKd7


freearhey (approved):
Great update! Thank you for your help!

PR #883, opened 2020-01-05
Update cz.m3u

lopapata:
Add "RTM plus" channel stream 
and changed name of Óčko Gold to Óčko STAR which is official name after rebrand.


freearhey:
Thank you for your help!

PR #924, opened 2020-01-26
Create us1.m3u

rford91895:
Created custom list of channels to include oxygen and other watched channels.


adamo331 (commented):
Please don't add like this list, they all test list or hacked list from provider IPTV and most them available in forum everywhere. all channels off

PR #945, opened 2020-02-04
Update ae.m3u

MoonSyria:
I updated many broken links

PR #949, opened 2020-02-05
Set ae.m3u channels language to Arabic

Link4K:
So they show up in https://iptv-org.github.io/iptv/index.language.m3u.

I think https://github.com/iptv-org/iptv/pull/945 removed them


freearhey:
Yeah, you're right. Thank you for your help!

PR #1066, opened 2020-03-21
Update de.m3u

Nintendocustom:


freearhey:
Could you explain why you want to delete these channels?

PR #1063, opened 2020-03-21
Update id.m3u

mikhaelpurba:
Delete some broken channels, adding some info, adding TVRI network on Java.

PR #1074, opened 2020-03-23
Add files via upload

data14a:


freearhey:
Sorry, but this makes it even more complicated. I'm gonna try to explain it more clearly. 

When you make any changes to one of the existing files and make a normal pull request, we can immediately see what you have changed, how different it is from the original and if necessary check that everything is working fine. The other situation is when you upload a new file from your computer. In this case we see only the content of the file itself. 

Here, you can compare for yourself:
https://github.com/iptv-org/iptv/pull/1063/files (one of the previous pull requests)
https://github.com/iptv-org/iptv/pull/1074/files (your file)

So it would be great if you edit the current version of the playlist and just create a new pull request. And there's also a guide for that: https://help.github.com/en/github/managing-files-in-a-repository/editing-files-in-your-repository

PR #1086, opened 2020-03-27
Added several Portuguese TV channels

ksl1989:
* Added the following Portuguese channels:
  * RTP2
  * RTP3
  * RTP Internacional
  * RTP Memória
  * RTP Madeira
  * RTP Açores
  * RTP África
  (Public broadcasting station channels in the qualities available at the official site)
  * ARTV (parliament channel)
  * Porto Canal (private channel owned by FC Porto football club)
* Removed the link for RTP1 and added the one from the official page.

A couple notes about the RTP streams:
* These are the streams used in the official page (https://www.rtp.pt/play/), they're fast, have good image quality and are stable but there are two issues:
  * When any of these channels is broadcasting a programme it doesn't have the rights to broadcast online (usually that's imported programmes) the m3u file will be empty, it may seem the stream is broken but it'll continue streaming as soon as this programme ends.
  * It doesn't work with Kodi's user-agent (and probably most IPTV apps' user-agents), but all streams are available just using "Mozilla" as a user-agent. I tried all links on Kodi for a couple days and they're all working. Maybe it should be noted on the README file that some streams may not work with the default user-agent.
* The RTP1 stream that was on the list was working but I think it's better to have the official good quality stream rather than this low quality stream of an unknown source.
* I used the `tvg-id` and `tvg-name` from https://siptv.eu/codes/ and https://www.iptv-epg.com/available-channels as these seem the most standard, this way it's more likely to work with most EPG sources.

### DIFF
--- a/channels/pt.m3u
+++ b/channels/pt.m3u
@@ -17,8 +17,6 @@ http://c2.manasat.com:1935/kmusic/kmusic3/FluxusTV.m3u8?fluxustv.m3u8
 http://195.22.11.11:1935/ktv/ktv1/FluxusTV.m3u8?fluxustv.m3u8
 #EXTINF:-1 tvg-id="" tvg-name="" tvg-language="Portuguese" tvg-logo="https://i.imgur.com/M5i4hfh.png" group-title="",Kuriakos TV
 http://195.22.11.11:1935/ktv/ktv1/TeamBlue.m3u8
-#EXTINF:-1 tvg-id="RTP 1 PT" tvg-name="RTP 1 PT" tvg-language="Portuguese" tvg-logo="https://i.imgur.com/Rdt24TC.png" group-title="",RTP 1
-http://210.210.155.35/qwr9ew/s/s38/index.m3u8
 #EXTINF:-1 tvg-id="" tvg-name="" tvg-language="Portuguese" tvg-logo="https://i.imgur.com/jYIf7Mz.png" group-title="",Sobrenatural TV
 http://213.13.26.11:1935/live/sobrenaturaltv/livestream.m3u8
 #EXTINF:-1 tvg-id="" tvg-name="" tvg-language="Portuguese" tvg-logo="https://i.imgur.com/VgzVZto.png" group-title="",Sobrenatural TV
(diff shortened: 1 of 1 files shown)


freearhey (changes requested):
You have done a great job, but in order for us to approve this pull request you will need to make some changes:
1. We specifically add multiple broadcast sources for the same channel as a backup in case one of the options stops working. So the link to RTP1 will need to be returned.
2. If the broadcast is interrupted periodically, you can specify this in the channel name within square brackets. The same applies to broadcasts that do not work in certain applications.
3. `tvg-id` must match the ID from the EPG file specified in the playlist header, in this case http://195.154.221.171/epg/guideportugal.xml.gz Also, the `tvg-name` must match the `display-name` from this EPG. This way, others can use this EPG with these channels without any additional configuration.

ksl1989:
1. I think right now the problem is there are too many repeated channels with the same name. There should be no repeated channel unless it's the same channel in different quality (or some other explicit difference). There are channels repeated dozens of times, that makes it more confusing and harder to understand. I saw some Chinese channels repeated dozens of times with the same name, that makes it very hard to navigate and know what to choose. Just one backup is not too bad, though. Do you mind appending "(Backup)" or something similar to the name, just so there are no repeated names? Personally I think I would be better to keep a separate list of backup streams to be substituted if one stops working. 
2. 
  * "If the broadcast is interrupted periodically, you can specify this in the channel name within square brackets." — if think that's a good idea, but exactly what do you suggest writing? "Interrupted periodically" sounds like it goes on and off for no reason, that's not the case.
  * "The same applies to broadcasts that do not work in certain applications" it should work in most applications as most allow for changing the user-agent and this should be relatively common. We could also add "#EXTVLCOPT:http-user-agent" for VLC and/or append "|user-agent=" for Kodi but I was trying to leave it player agnostic. Again, what exactly do you think should be written? I agree it should be noted, I just don't think it should be part of the channel name.
3. I noticed the name matched the name on the EPG file but it wasn't working for me (and it is for other channels, so it's probably not a config issue) and since there was nothing to lose I thought it was better to go with more standard names. I can check the file and change the channels it has to match the name ID there.

isblue01:
agree

freearhey:
@ksl1989 
1. I think adding "(Buckup)" to the channel name is a good choice.
As for the idea to create a separate list for backup links, I think there is common sense in that. But since this will require updating all playlists, I think we should create a separate issue for this and get the opinion of the rest of the community.
2. 
- Some people use "[not 24/7]" for these channels, which I think is a good choice. 
- As for using "#EXTVLCOPT" for `user-agent`, I think it's a really good idea. Also I just checked and it turns out that `#EXTVLCOPT` works fine not only in VLC player but also in Kodi. So we can safely use this attribute instead of specifying it in the channel name. The only thing is that it will take me some time to add support for this attribute to the playlist update script, otherwise it will just be erased from the final version.
3. If `tvg-id` and `tvg-name` are specified correctly then EPG will display normally in all applications that support it. In your case, none of the links contain the corresponding `tvg-id` or `tvg-name`. If some particular channel is missing in this EPG file, it is better to just leave the `tvg-id` and `tvg-name` attributes empty.

adamo331 (commented):
all the streams link portugal not working in VLC also not working in official website

freearhey:
@adamo331 I just ran RTP1 through the site (https://www.rtp.pt/play/direto/rtp1) and everything is working fine fo me. 

It is true that not all channels are working at the moment, but @ksl1989 mentioned earlier that not all channels are working twenty-four hours a day.

ksl1989:
> I think adding "(Buckup)" to the channel name is a good choice.

All right, I'll do that then.

> As for the idea to create a separate list for backup links, I think there is common sense in that. But since this will require updating all playlists, I think we should create a separate issue for this and get the opinion of the rest of the community.

I'm not telling you to do this, just suggesting this would be more sensible going forward. Some countries' lists will need some serious cleanup; the Chinese one especially really has some channels repeated dozens of times. The same way you have an unsorted list, you can have a backup list.

> Some people use "[not 24/7]" for these channels, which I think is a good choice.

I don't think it is a good choice because it gives the idea the channel is just open for a few hours a day but actually some of those channels have very few imported shows and can actually be streaming 24/7 for most of the time. But since I can't think of something short that describes this situation, "[not 24/7]" can work for now. "Occasionally interrupted due to broadcast rights uses" would be right but it's too long.

> As for using "#EXTVLCOPT" for user-agent, I think it's a really good idea. Also I just checked and it turns out that #EXTVLCOPT works fine not only in VLC player but also in Kodi. So we can safely use this attribute instead of specifying it in the channel name. The only thing is that it will take me some time to add support for this attribute to the playlist update script, otherwise it will just be erased from the final version.

I was convinced it wasn't supported in Kodi, so I looked it up again and it turns out Kodi doesn't support it natively (if you open the m3u playlist directly it doesn't work) but the IPTV Simple PVR  addon does support it (I tested and confirmed this), and that's likely how most people will use this list anyway. Other players may support it or not. If you confirm this is acceptable to you, I'll do it like this.

> If tvg-id and tvg-name are specified correctly then EPG will display normally in all applications that support it. In your case, none of the links contain the corresponding tvg-id or tvg-name.

Actually IPTV Simple PVR [does't support](https://github.com/kodi-pvr/pvr.iptvsimple/issues/308) support setting the EPG file in each individual stream, only at the file level with `x-tvg-url` and because you are using the names and IDs from this specific file, it'll only work if you set it separately in the add-on settings. If you use the general file with all the links, you'll need to pick which EPG file will be read, only one can be used. Does any application actually respect the EPG set in this way?

> If some particular channel is missing in this EPG file, it is better to just leave the tvg-id and tvg-name attributes empty.

If we do this we prevent the user from using any EPG file, even if the have one with the correct information and can correct the ID and name. If no ID and name are set, there is nothing to match. I still think it's better the try to use the most standard name and ID in this case. Anything is better than nothing; something is usable, nothing isn't. And the more standard the more likely to work for most people.

ksl1989:
@adamo331 
> all the streams link portugal not working in VLC also not working in official website

ARTV and Porto canal links are working 24/7, RTP channels work when not broadcasting programmes for which they don't have the online broadcasting rights.

You're right about VLC, I also couldn't get it to play the RTP streams, but this seems to be a bug, if you set the user-agent to Mozilla it can access the m3u8 file and read the ts files but fails with a "Failed to create demuxer (nil) TLS" error. I see this bug has been mentioned [elsewhere](https://superuser.com/questions/1379361/vlc-and-m3u8-file).

channels/pt.m3u
```diff
@@ -17,8 +17,6 @@ http://c2.manasat.com:1935/kmusic/kmusic3/FluxusTV.m3u8?fluxustv.m3u8
 http://195.22.11.11:1935/ktv/ktv1/FluxusTV.m3u8?fluxustv.m3u8
 #EXTINF:-1 tvg-id="" tvg-name="" tvg-language="Portuguese" tvg-logo="https://i.imgur.com/M5i4hfh.png" group-title="",Kuriakos TV
 http://195.22.11.11:1935/ktv/ktv1/TeamBlue.m3u8
-#EXTINF:-1 tvg-id="RTP 1 PT" tvg-name="RTP 1 PT" tvg-language="Portuguese" tvg-logo="https://i.imgur.com/Rdt24TC.png" group-title="",RTP 1
-http://210.210.155.35/qwr9ew/s/s38/index.m3u8
```
freearhey, line 21: The link works, so please bring it back.

freearhey (changes requested):

channels/pt.m3u
ksl1989: The link works but it's mislabled, it turns out this is RTP Internacional, not RTP1. I didn't realize it at first because the channels share most pf their programming but I'll correct it and update the PR. There are a number of other changes to make here as well.

ksl1989:
I was waiting for you to confirm if you'd accept a PR with `#EXTVLCOPT` lines for the user-agent issue and I still don't agree with leaving the `tvg-id` attrbute empty, which can prevent people from using even a custom EPG source, but  I did as requested.

* Changed RTP links to multi-quality playlist rather than individual bitrate streams.
  (I hadn't realised all channels have a mult-bitrate playlist, this is better, although most players don't make changing bitrates easy.)
* Added #EXTVLCOPT line to channels that need the user-agent to be set.
* Changed tvg-id and tvg-name to the one in provided EPG file.
* Re-added removed RTP1 stream as backup for RTP Internacional (was mislabled).
* Added "[not 24/7]" to the title of RTP channels because the broadcast can be interrupted due to rights issues
  (I think at least RTP África and RTP Internacional may broadcast 24/7 but I can't be sure and the only way to confirm would be watching 24/7. Maybe also Madeira and Açores.)

freearhey (approved):

PR #1099, opened 2020-03-29
Updated int.m3u

themak1985:
Removed Broken Links, Added Missing Languages, fixed incorrect languages.
A lot of those vietnamese channels are online only channels - my guess is that they should be moved to vn.m3u but I wanted to do this first.

### DIFF
--- a/channels/int.m3u
+++ b/channels/int.m3u
@@ -1,5 +1,5 @@
 #EXTM3U
-#EXTINF:-1 tvg-id="" tvg-name="" tvg-language="English" tvg-logo="https://i.imgur.com/HGmC8ea.png" group-title="",24 Box TV
+#EXTINF:-1 tvg-id="" tvg-name="" tvg-language="Persian" tvg-logo="https://i.imgur.com/HGmC8ea.png" group-title="",24 Box TV
 http://51.158.190.97/hls/stream.m3u8
 #EXTINF:-1 tvg-id="" tvg-name="" tvg-language="English" tvg-logo="https://i.imgur.com/HL7fwzt.png" group-title="",Afrobeats
 https://stream.ecable.tv/afrobeats/tracks-v1a1/mono.m3u8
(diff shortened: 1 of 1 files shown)


channels/int.m3u
```diff
@@ -27,33 +27,19 @@ https://www.ast.tv/stream/2/ultra.m3u8
...
-#EXTINF:-1 tvg-id="" tvg-name="" tvg-language="" tvg-logo="" group-title="XXX",Big Tits HD
-http://nruxmzi.ojswi5dsmftgm2ldfz4hs6q.cmle.ru/bigtits.m3u8
-#EXTINF:-1 tvg-id="" tvg-name="" tvg-language="" tvg-logo="" group-title="XXX",BIGASS
-http://nruxmzi.ojswi5dsmftgm2ldfz4hs6q.cmle.ru/bigass.m3u8
-#EXTINF:-1 tvg-id="" tvg-name="" tvg-language="" tvg-logo="" group-title="XXX",BIGDICK
-http://nruxmzi.ojswi5dsmftgm2ldfz4hs6q.cmle.ru/bigdick.m3u8
-#EXTINF:-1 tvg-id="" tvg-name="" tvg-language="" tvg-logo="" group-title="XXX",BLONDE
-http://nruxmzi.ojswi5dsmftgm2ldfz4hs6q.cmle.ru/blonde.m3u8
-#EXTINF:-1 tvg-id="" tvg-name="" tvg-language="" tvg-logo="" group-title="XXX",BLOWJOB
-http://nruxmzi.ojswi5dsmftgm2ldfz4hs6q.cmle.ru/blowjob.m3u8
```
freearhey: Could you explain why you deleted those links?

freearhey (changes requested):

freearhey:
As for the Vietnamese channels, I agree, it would be better to move them to the appropriate playlist.

channels/int.m3u
themak1985: Did they work for you? I couldn't get them to work - even when I used streamtest.in

themak1985: Looks like they are geoblocked. I just changed my vpn to Ukraine and it worked - US, UK, Slovakia, South Africa, all failed. I don't have a russian VPN to check if that could work but based on the .ru address - I will assume yes.

freearhey: You're right, links only work from some locations.

PR #1194, opened 2020-04-16
Added several lusophone channels

ksl1989:
* Angola
  * Added TPA1 (Public broadcaster)
* Mozambique
  *  Added TVM (state broadcaster)
  * Added TVM Internacional
  * Added Miramar
  * Added category to Trace Toca e TV Maná Moçambique

Note: I'm not sure Trace Toca should be here, it seems it's owned by an international company and broadcast in all Portuguese-speaking Africa, not just Mozambique

* Cape Verde
  * Replaced link with the one from the official page (the one that was there wasn't working for me)
* Spain
  * Corrected the language — Televisión de Galicia broadcasts in Galician (sometimes considered a variant of Portuguese), not Spanish.
  * Added category
  * Removed repeated channel (medium quality stream already included in multi-quality playlist below).
  * Changed the name as both links are two versions of the same channel (European and American). I'm using the full name as it appears in the official website.

Note: There are two livestream.com links. These links are probably not permanent but these are from the official website, and should be relatively stable.

### DIFF
--- a/channels/ao.m3u
+++ b/channels/ao.m3u
@@ -1,1 +1,3 @@
 #EXTM3U
+EXTINF:-1 tvg-id="" tvg-name="" tvg-language="Portuguese" tvg-logo="https://i.imgur.com/12aPxO7.png" group-title="General",TPA1
+https://api.new.livestream.com/accounts/1181452/events/8865379/live.m3u8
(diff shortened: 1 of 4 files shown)


freearhey:
If you are not sure which playlist a particular channel should be on, you can use the country specified in the channel description at [lyngsat.com](https://www.lyngsat.com/tvchannels/fr/Trace-Toca.html)

ksl1989:
Thanks, that's a good tip, but in this case I'm still in doubt. It says France because the company is probably French, at least legally, but it also shows the channel being offered only by Portuguese and African providers. This is a Portuguese(-language) African music channel, it would probably be better placed in the international playlist. I'm not sure so I'll leave it where it is for now.

freearhey:
And we also have an international playlist: https://github.com/iptv-org/iptv/blob/master/channels/int.m3u

PR #1203, opened 2020-04-17
Added TVI channels

ksl1989:
* Added TVI channels (TVI, TVI24, TVI Reality, TVI Internacional, TVI Ficção)

These channels are available publicly on [TVI's website](https://tviplayer.iol.pt/direto) but require an expiring token, also publicly generated, so I made a [simple script](https://github.com/ksl1989/static-m3u8/blob/master/index.php) to have static links be redirected to the dynamically generated link (URL+token).

It would be great if somebody here had a bit of server space to host this script, bandwidth usage is minimal, it literally just gets the link (or token) and redirects to the correct link. Right know I'm hosting it on  a free hosting provider but I don't know if they'll allow it to stay, free hosting providers don't like their server space used for scripts. For now it's working fine, it's not even slow, but I don't know how long it'll last.

The idea is to eventually add other similar links so these public links can also be added to m3u playlists, it's such a loss if we can't use all of these public links just because they require a (public) token or API. You will notice I also have all the Euronews channels from their public API on this script but it's not working from this server, there's an SSL setting that needs to be changed and I can't do it in a free shared server.


freearhey:
I think that's a good idea! But if you want most of the community to see your message, it's best to create a separate issue for that.

PR #1219, opened 2020-04-19
Update uk.m3u

kimbrasil:


nepalicana:
Good to see so many channels but most of them doesn’t work for me and gives me “OOPS SORRY, LIMIT REACHED”. Are there anyone facing the same problem?

kimbrasil:
> Good to see so many channels but most of them doesn’t work for me and gives me “OOPS SORRY, LIMIT REACHED”. Are there anyone facing the same problem?

Seems that most of them are really down now, I'd used them for days. So, I'll try to pull only the URLs that already works. TY for advice!

SileC:
They work until they get public then either you get message oops reach limit or cannot open.

PR #1281, opened 2020-04-26
Update be.m3u

Sphinxroot:
added : Matele , rtc liege , tv lux , vedia


freearhey (commented):
tvg-id must be the same as channel ID in EPG file

PR #1364, opened 2020-05-11
Update Reshet13 stream in il.m3u8

tomer:
Previous URL returns 404

Also, I'd suggest to re-arrange the streams in the file, as the primary channels are mixed with channels with very low tv rating.


kimbrasil:
I suggest change the URL to the chunklist, available at https://reshet-live-il.ctedgecdn.net/13tv-desktop/r13.m3u8

tomer:
> I suggest change the URL to the chunklist, available at https://reshet-live-il.ctedgecdn.net/13tv-desktop/r13.m3u8

Good idea. I've updated the URL.

freearhey (approved):

freearhey:
All playlists are automatically sorted alphabetically because this is the only way we can maintain order, not only in this playlist, but also in the other hundred playlists.

PR #1522, opened 2020-06-16
Update bd.m3u

ghost:
i updated this with all channels i have.some of them only works with bd ip


freearhey:
Unfortunately, I cannot approve this pull request.

First of all, we do not add geo-blocked channels to playlists so it makes it very difficult to support them further.

Secondly, the channels in playlists are sorted by country of broadcast source and not by country of broadcast available for viewing. In your list most of the channels are not from Bangladesh.

To avoid these mistakes in the future, please read this guide before making any changes: https://github.com/iptv-org/iptv/blob/master/.github/CONTRIBUTING.md#add-channel

badhon495:
brother i updated again.
now i removed non Bengali channel.
and they are not area restricted now.
thanks :)

PR #1560, opened 2020-07-03
Update in.m3u

emlimap:
Marutam Music doesn't seem to have a logo on their website that we could use so just used Marutam TV's logo.


freearhey (approved):

freearhey:
If you can't find the corresponding channel logo, just leave the field blank.

PR #1574, opened 2020-07-07
Added missing tvg-language to gr.m3u

themak1985:


freearhey:
There is a problem: the ISO 639-3 standard that we use in the project does not have such a language as "Greek" only "Modern Greek (1453-)" - https://iso639-3.sil.org/code_tables/639/data?title=&field_iso639_cd_st_mmbrshp_639_1_tid=All&name_3=Greek&field_iso639_element_scope_tid=All&field_iso639_language_type_tid=All&items_per_page=200

And even if I approve this PR language will be erased the next time you update your playlist automatically.

themak1985:
Went ahead and changed from Greek to Modern Greek. (Less user friendly imho, but its what the standard says)

freearhey:
It also won't work without the " (1453-)".

And yes, I think the first version is better. But I also think if we follow a standard, we should follow it completely.

freearhey (approved):

PR #1657, opened 2020-08-05
[Update] & [First warning] - No bit.ly links allowed. If we catch you will be banned from IPTV

I-Cat:
Please note adding an ad / paid to click on / url like this can cause BIG problems. This Is Your First warning to Whom ever added this. If We see it again. You will be banned from IPTV. Thankyou for understanding.


I-Cat (commented):
ok

freearhey:
I cannot approve this PR because the files that are automatically generated are changed.

PR #1658, opened 2020-08-05
Patch 2 Removed ad based links aka bit.ly

I-Cat:
Removed ad based links aka bit.ly


I-Cat:
Fix

freearhey:
I cannot approve this PR because the files that are automatically generated are changed.

PR #1737, opened 2020-09-14
Update ae.m3u

jad222:
Add mbc1


jad222:
can you update json and m3u files ? i can't get the channels i've added yesterday !
@freearhey

freearhey (approved):

freearhey:
All playlists are automatically updated every day at 00:00 GMT. So tomorrow all added channels should appear everywhere.

PR #1749, opened 2020-09-16
Update SY

jad222:
Better quality


freearhey (changes requested):
We store all versions of the stream just in case.

jad222:
Ok , you can add it like another stream option :)

PR #1766, opened 2020-10-27
Add legal section to readme

Daniel15:
Hopefully this will make people actually think about their actions before sending a DMCA.


freearhey (approved):

freearhey:
Let's hope so, although GitHub's position to block the repository without any warning still doesn't give me much confidence in this matter.

PR #1781, opened 2020-10-30
Adds public service channels from DR

andreasmalling:
This commit adds the 3 biggest channels from Denmarks Radio (DR) providing public service in Denmark


freearhey:
Links no longer work.

andreasmalling:
> Links no longer work.

They are still working in Denmark. There seems to be geo blocking on them. What is your stance on this? These channels are still highly relevant for people living in Denmark.

freearhey:
We do not add geoblocked links to playlists as they are too difficult to maintain.

PR #1789, opened 2020-11-08
Update CA streams

tpkahlon:
- Remove duplicates
- Remove broken streams
- Add CBC, Global, ICI Tele streams


freearhey:
Links no longer work.

tpkahlon:
@freearhey Which links no longer work? Let me know, I can fix them. When I tested, I checked each link carefully.

freearhey:
At the moment all of them do not work for me.
Screenshot 2020-11-12 at 13 44 21

tpkahlon:
See my [screen-cast](https://imgur.com/9k1Y5Oi) for reference. Seems to work on my end. Please try to use [Licecap](https://www.cockos.com/licecap/) on your end so you can walk me through the bug. Screenshot does not help much.

freearhey:
This only means that broadcasts are geoblocked by the provider and we do not add such broadcasts to the playlist to avoid same problems in the future.

tpkahlon:
Interesting. I did not know that. Is there a way to validate/test if a stream is geoblocked?

freearhey:
I usually use https://streamtest.in/ to check the availability from other places, but I think there are many other similar services.

PR #1799, opened 2020-11-16
✨ Create SFW versions of the playlists

AnotherFoxGuy:
The generate.js script can now also generate SFW versions of the playlists (it will filter adult channels out)

Changes:
- Updated `npm run generate`, this will now generate the normal playlists and SFW playlists
- Added `npm run generate:full`, this will generate the normal playlists 
- Added `npm run generate:sfw`, this will generate playlists with a `.sfw` prefix without any adult channels

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@ node_modules/
 error.log
 helpers/categories.md
 helpers/countries.md
+.gh-pages
 
 # IDE
 *.sublime-project
--- a/.readme/template.md
+++ b/.readme/template.md
@@ -6,7 +6,8 @@ Internet Protocol television (IPTV) is the delivery of television content over I
 
 ## Usage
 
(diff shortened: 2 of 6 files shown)


wenye618 (approved):

salamwaddah:
merge

LaneSh4d0w:
Well, for now, I think it's better to wait for every channel to be grouped correctly (and for the unsorted channels to be sorted) because they are, for now, 3k non grouped channels, several of them being NSFW. Taking this idea in general, it's sorta non-productive, because some NSFW channels will be on the SFW playlist (unsorted or ungrouped).

If you took those who were grouped, you would face an incredibely low number of channels.

It LGTM, but won't be useful for the moment, we need to sort all channels (and it will take some time).

freearhey:
As @LaneSh4d0w correctly pointed out, we cannot be sure that the playlist is SFW if we just filter out all the channels labeled XXX. 

So for now I'm closing this PR.

Issue #1831, opened 2020-12-17
DMCA Takedown Notice


freearhey:
Today I received the following email from GitHub Support:

>Hi IPTV,
>
>I'm contacting you on behalf of GitHub because we've received a DMCA takedown notice regarding the following content:
>
>https://github.com/iptv-org/iptv
>
>We're giving you 1 business day to make the changes identified in the following notice:
>
>https://github.zendesk.com/attachments/token/k9lfsiwWTMg1CkAPpQ85GvzM7/?name=2020-12-17-Nagravision.rtf
>
>If you need to remove specific content from your repository, simply making the repository private or deleting it via a commit won't resolve the alleged infringement. Instead, you must follow these instructions to remove the content from your repository's history, even if you don't think it's sensitive:
>
>https://docs.github.com/articles/remove-sensitive-data
>
>Once you've made changes, please reply to this message and let us know. If you don't tell us that you've made changes within the next 1 business day, we'll need to disable the entire repository according to our GitHub DMCA Takedown Policy:
>
>https://docs.github.com/articles/dmca-takedown-policy/
>
>If you believe your content on GitHub was mistakenly disabled by a DMCA takedown request, you have the right to contest the takedown by submitting a counter notice, as described in our DMCA Takedown Policy.
>
>PLEASE NOTE: It is important that you reply to this message within 1 business day to tell us whether you've made changes. If you do not, the repository will be disabled.

The `2020-12-17-Nagravision.rtf` file linked to in the email contains the following:

>NAGRA Notice ID: DISC-20201115NEB36461
>Notice Date: November 15, 2020
>URL: https://iptv-org.github.io/iptv/categories/documentary.m3u
>Source: https://iptv-org.github.io/iptv/categories/documentary.m3u
>Domain: iptv-org.github.io
>
>Dear Sir, Madam,
>
[...]

PR #1832, opened 2020-12-17
Remove Discovery channels

freearhey:
Remove links to all channels mentioned in the Nagravision DMCA Notice from 2020-12-17.

Details: #1831


Daniel15 (approved):
Thanks! I hope the copyright holders understand that they need to send a DMCA request to the actual hosts of the content, rather than just this repo. Removing them is a good idea though.

themak1985 (approved):
Oddly enough, I checked most of the links and they are not working anyways. 
@freearhey I'll follow your lead if this is normal protocol to adhere to copyright claims, even though I agree with @Daniel15 - we are not violating copyright. 
Your changes are good.

freearhey:
@themak1985 The point is that this is the second complaint since the repository was created, and the last time there were no warnings at all, so unfortunately there is no guarantee that the repository will not be blocked anyway.

PR #1841, opened 2020-12-21
Add Arabic EPG

link8k:


themak1985:
@link8k thanks for finding an epg that works for 12 countries. Before approving, can you ensure that each impacted country's playlist includes the tvg source? Changing top line from:
`#EXTM3U` 
to
`#EXTM3U x-tvg-url="https://raw.githubusercontent.com/Fazzani/grab/master/merge.zip"`

We would also want to make sure tvg-id for all applicable channels is updated, but that can happen at a later date.

link8k:
Thanks, The tvg-id can be updated eventually.

PR #1850, opened 2020-12-26
Update so.m3u

link8k:


themak1985:
all links tested, thanks for adding @link8k

PR #1852, opened 2020-12-27
Remove old North Korea link from South Korea

LmaoLover:
Korean Central Television is in the kp.m3u list (North Korea)

This link is also outdated and does not work.


themak1985:
Link is dead - merged.

PR #1853, opened 2020-12-27
Add CTV Coptic Orthodox Church Channel

PolaEdward:
This is one of the official channels of the Egyptian church broadcasting 24/7 christian arabic content, and sometimes broadcast in english and coptic.


themak1985:
Thanks for the addition @PolaEdward - added.

PR #1855, opened 2020-12-27
Add Aghapy TV channel

PolaEdward:


themak1985:
Also note, if the channel is not in the guide source, you can just leave tvg-id blank.

PR #1871, opened 2020-12-29
Spain iran unsorted

themak1985:


themak1985:
@freearhey I went through unsorted and there were about 90 or so streams from Spain. After checking them all, most were working, however all but one were radio stations with audio only feeds. I moved the one that was an actual tv channel.  I assume we should not include the audio only feeds, but I will include it if you think we should.

freearhey:
@themak1985 I totally agree with you. Most likely they got into this playlist by accident along with the rest of the links. So you can safely remove all streams that contain only the audio track.

PR #1931, opened 2021-01-12
Update int.m3u

reklamalinir:


LaneSh4d0w:
Highly sounds like a ripoff from RedTraffic's livestreams. Could be used as backups.

themak1985:
@LaneSh4d0w I actually think the RedTraffic livestreams are backups of these. @reklamalinir do you know how these relate to red traffic?

LaneSh4d0w:
> @LaneSh4d0w I actually think the RedTraffic livestreams are backups of these. @reklamalinir do you know how these relate to red traffic?

It could be, to be honest I've seen more RedTraffic streams than AdultIPTV, and I never seen the localisation of the servers.

freearhey (approved):

PR #1998, opened 2021-01-17
Clean playlists (and move unsorted data) - P2

BellezaEmporium:
This is a follow-up to what I was doing over time.

As always, reviews are appreciated.


LaneSh4d0w:
Potential answers to some questions : 

HBO Asia is made by HBO, which is located in the US. Since we sort the channels by the country the channel is originated from, that's why HBO Asia is in the US.

HDFashion & Lifestyle isn't an US channel, it's an Ukrainian one : http://hdfashion.tv/contacts

Lots of dupes were found on the unsorted m3u file, which could explain things about why lots of channels are deleted.

Yes, some Chinese channels got their token lost, because it isn't needed for the channel to work (tested and deleted when working without it)

LaneSh4d0w:
Ready for review.

freearhey (approved):

PR #2016, opened 2021-01-19
Update ir.m3u

data14a:
Added new channels from Persiana group and removed channels that no longer available.

### DIFF
--- a/channels/ir.m3u
+++ b/channels/ir.m3u
@@ -119,6 +119,30 @@ http://iptv.tapesh.tv/tapesh/playlist.m3u8
 http://iptv.tapesh.tv/tapesh/playlist1/index.m3u8
 #EXTINF:-1 tvg-id="" tvg-name="Persian Bazar" tvg-language="Persian" tvg-logo="https://www.parsatv.com/index_files/channels/persianbazar.png" group-title="",Persian Bazar
 https://stream.persiantv1.com/ptv1/playlist1/index.m3u8
+#EXTINF:-1 tvg-id="" tvg-name="Persiana BILLBOARD" tvg-language="Persian" tvg-logo="http://persianagroup.tv/wp-content/uploads/2020/08/Persiana-Billboard.png" group-title="Music",Persiana BILLBOARD
+http://51.210.199.10/hls/stream.m3u8
+#EXTINF:-1 tvg-id="" tvg-name="Persiana CINEMA" tvg-language="Persian" tvg-logo="https://persianagroup.tv/wp-content/uploads/2020/08/Persiana-Cinema.png" group-title="Movies",Persiana CINEMA
+http://51.210.199.15/hls/stream.m3u8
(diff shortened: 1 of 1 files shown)


LaneSh4d0w:
Persiana Group = French company

Links are already available in the French playlist.

data14a:
They are Persian speaking channels and they should not be int the French playlist.

LaneSh4d0w:
As @freearhey said, we do not classify the channels on their spoken language, but from where do they come from. Persiana being a French company, it's in the French playlist.

data14a:
Because of the political climate in Iran, majority of the TV production is outside Iran. In Iran's playlist there are only 12 channels that are actually broadcast from Iran.

LaneSh4d0w:
That for sure is something Interesting to know. What are they, if I may ask ? We will relocate them to their correct location.

data14a:
I have tried to contribute to this project but is getting impossible if each time the request is denied because the channel is not from Iran. Iran is a country of multiple languages including: Persian (Farsi), Arabic, Kurdish, Azerbaijani and some other languages. There are no private TV stations in Iran. TV and radio stations are run by the government of Iran (IRIB), So any other channels is broadcasted from outside of Iran (US, UK, France, etc.) targeting the people of Iran. It would make sense to keep this TV station Under the country of Iran regardless of where they are located.

LaneSh4d0w:
@freearhey thoughts ?

freearhey:
@data14a My advice would be before you add a channel just check what country is listed in [lyngsat.com](https://www.lyngsat.com/search.html).

data14a:
As I explained before, almost are Iranian channels are operated outside Iran. So, listing Iranian (Persian) channels under France, US or UK does not make any sense because they are not intended for people in those countries.

freearhey:
@data14a I understand, but you also understand me, you just state the problem and do not give any solution. In addition to the Iranian playlist there are almost 200 other playlists and we have to somehow keep them in order. 

If you have an idea how to better organize all 5000 channels then please post your proposal [here](https://github.com/iptv-org/iptv/issues/new?assignees=&labels=discussion&template=feature-request.md&title=Propose%3A+xxx) first, so that the rest of the community can give their opinion as well.

data14a:
I just do not understand the inconsistency in applying the rules. There are 64 channels under country of Iran, out of those only 16 (some repeat channels) are broadcasted from Iran. The other 48 from multiple countries.

freearhey:
@data14a Which ones?

data14a:
Any channel that is not labeled IRIB.

LaneSh4d0w:
> @data14a Which ones?

Maybe we should classify channels in their language...

data14a:
I do not know the right path for the entire site but personally I would like to keep all the channels that are intended for people of Iran under the same m3u list regardless of the language of the TV station.

PR #2034, opened 2021-01-22
Fix hibridcdn.net channels

link8k:


LaneSh4d0w:
They completely changed their server at the same instance. chunks.m3u8 also is an adaptive playlist, it's just a different name.

link8k:
No, it was only a 1080p stream. This one now has different qualities.

LaneSh4d0w:
> No, it was only a 1080p stream. This one now has different qualities.

Nice find.

LGTM

freearhey (approved):

PR #2138, opened 2021-02-11
change regions

themak1985:


themak1985:
A lot of channels broadcast areas are very wide, for example 200 channels from arab countries broadcast to diaspora communities in the US called SlingTV. I don't think that justifies including the US as a country. The Abu Dhabi channels are an example of that. 
Peace TV tries to claim they serve all those regions, but they are actually banned in a bunch of countries and mostly just stream online. I placed as INT as a placeholder.

freearhey (approved):

PR #2145, opened 2021-02-13
Redefine APAC Region

themak1985:


themak1985:
Justification: Per the wiki article we link to: Western and Central Asia are almost never included in this region. I suggest we remove them. I kept Russia, because northern Asia which only consists of Siberia is technically a part of APAC, but some could argue otherwise and to remove it. So I left RU alone.

freearhey:
You are right, Russia should not be included in this region.

freearhey (approved):

PR #2176, opened 2021-02-22
Add dropdown to reademe template

themak1985:
The contribution section gets lost all the way at the bottom, suggestion to place right before usage. Alternatively, right after usage but before playlists by category, but we would have to change some of the language at the end of usage. Another alternative is to keep where it is but create anchor links at start of readme.

### DIFF
--- a/.readme/template.md
+++ b/.readme/template.md
@@ -18,7 +18,7 @@ Also you can instead use one of these playlists:
 
 Or select one of the playlists from the list below.
 
-### By Category
+### By category
 
 <details>
 <summary>Expand</summary>
@@ -29,7 +29,7 @@ Or select one of the playlists from the list below.
 
 </details>
 
-### By Language
+### By language
 
 <details>
 <summary>Expand</summary>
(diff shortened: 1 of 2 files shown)


LaneSh4d0w:
I do agree. It will permit newcomers that want to contribute to see it directly on the README. Else they wouldn't mind getting all the way to the bottom page to see how to.

freearhey:
I think that's unnecessary. After all, most people come here to copy the playlist, not to edit it, and they don't need to read the guide at all.

That said, I think we can try to hide the tables behind the [dropdown menu](https://gist.github.com/citrusui/07978f14b11adada364ff901e27c7f61) and maybe this way we can make it a little easier to find the guide for those who need it.

themak1985:
@freearhey I implemented the drop downs, though I am not sure I like it. What do you think?

freearhey:
@themak1985 I made a few changes, but overall I think it looks good. 

And just a reminder, in order to change the README.md content you need to edit the `.readme/template.md` file.

themak1985:
@freearhey, ready for your review.

freearhey (approved):

PR #2210, opened 2021-03-03
Update fr.m3u

pcccw:


freearhey (changes requested):
All added links do not work.

pcccw:
Look like the new link are geo-locked in Europe.

PR #2284, opened 2021-03-25
PL: Channels fixes - switch and remove old

Triloworld:
There was small issue about naming. 
Switch according to names. 
Removed not working.
Russian from Poland list removed.
Added few more that are open to broadcast and are public available


Rebwar27:
how to import mu3 file to link and download in iptv player

Triloworld:
> how to import mu3 file to link and download in iptv player

It's this link you think about: https://iptv-org.github.io/iptv/languages/pol.m3u
On your iptv player just import it. It will recognize as a list of channels. 
What iptv player from milions you talking about? You "stream" programs, and you can just download list of links there: https://github.com/iptv-org/iptv/blob/d75cb46e3a46458651dd1890d059628a5aff428a/channels/pl.m3u
This will be officiall if they accept proposed fix and changes.

freearhey:
@Triloworld Sorry, but I do not understand on what basis you have chosen `tvg-id` for channels?

PR #2289, opened 2021-03-26
Add TV8

link8k:
https://streamtest.in/logs/https-tv8-personamedia-tv-tv8hls-type-1-fmt-hls-file-m3u8-rend-v_r2


freearhey (approved):

pcccw:
This is not the actual TV8 stream. It is an advertisement that keeps looping.

freearhey:
@pcccw Thank you for bringing this to our attention! I've already removed the link from the playlist.

PR #2310, opened 2021-03-30
Move Some Unsorted Channels to US

themak1985:


themak1985:
I am not 100% sure if the spanish language channels should be US or LATAM or something else. 
Also the german language channels maybe country DE and CH, but I am also not 100% sure.

freearhey (approved):

PR #2323, opened 2021-04-02
Remove Audio Only Streams

themak1985:


Sphinxroot:
im wondering if @freearhey would be interressed in a news repo for world wide radio / audio stream like IPRADIO-ORG and CCTV CAM world Stream ?

freearhey:
@Sphinxroot I just don't have time to do any other projects at the moment. But I'm sure if you decide to create something like this people will support you.

PR #2396, opened 2021-04-11
IT : Fix links and add Geolock mention

BellezaEmporium:
EDIT: nevermind, option won't work. Those channels are geo-locked.


LaneSh4d0w:
An alternative is ok.ru links, but I don't know how to handle those.

freearhey (approved):

PR #2398, opened 2021-04-11
This is massive

AntiPontifex:
Apologize for the n00bness; again, still learning!


freearhey (approved):

freearhey:
@AntiPontifex I see how much useful information you have added to the channel descriptions and I would love to accept your submission, but I can't because of a very serious conflict between the current version of the file and yours. This is because the file has been updated several times by other contributors since you did the repository fork. To avoid this in the future, I can advise you to modify the file in small batches and send us a pull request more often. 

There are also some critical syntax errors in the file, for example every link should have its own description, otherwise it will simply be skipped: https://github.com/iptv-org/iptv/pull/2398/files#diff-98966c372a73123f473b186fdf03575c0a95532c2689db1fd509f8c9b5d35992R2871-R2874

In this situation, so that the work done is not lost, you can try to fix the discrepancy between the two versions of the file yourself by syncing your fork with our repo (here is a short instruction on how to do it https://stackoverflow.com/a/53538020) and then create a new pull request.

But anyway, thanks for trying to help!

PR #2426, opened 2021-04-18
Not a polish channel removal

yoge67855:
This is Russian channel not polish


freearhey:
The thing is that within the repository we sort the channels by the country from which they are broadcasting and as I know this channel is broadcasting from Poland: https://www.lyngsat.com/tvchannels/pl/Belsat-TV.html

PR #2427, opened 2021-04-18
TH : Add some channels

BellezaEmporium:


freearhey:
Both channels are from Thailand.
https://www.lyngsat.com/tvchannels/th/Mono-29.html
https://www.lyngsat.com/tvchannels/th/Channel-3.html

LaneSh4d0w:
Done

freearhey (approved):

PR #2469, opened 2021-04-24
Update unsorted.m3u

Sphinxroot:
added some channel since i dont know where to place them :)


BellezaEmporium (approved):
Normally just sticking with the playlist.m3u8 links should be OK, but good call on adding those. All working on my side. You get my approval.

freearhey (approved):

PR #2495, opened 2021-04-26
Updated some Malayalam language channel logos

naseef:
Updated some Malayalam language channel logos as the old logos were hosted in mhdtvworld[dot]com which is blocked by most ISPs in india


freearhey (approved):

freearhey:
@naseef Thanks!

PR #2519, opened 2021-04-29
Update mk.m3u

aleksandarsajdovski:
added sitel and kanal5


freearhey (changes requested):
Both links don't work.
Captura de pantalla 2021-04-29 a las 12 02 36

aleksandarsajdovski:
The urls are restricted only for IPs coming from MKD

freearhey:
This is strange, because I get HTTP error 404, which usually means that the link does not work at all. But if you are sure that the broadcast still works at least in Macedonia, just add `[Geo-blocked]` to the channel name and then we can add it to the playlist.

PR #2621, opened 2021-05-04
Remove duplicates

ghost:


freearhey:
None of the links are duplicates.

tilfizyawn:
ok, but I think the resolution should be part of the title when adding different resolution links.

freearhey:
You're right, and that's exactly what our [guide](https://github.com/iptv-org/iptv/blob/master/CONTRIBUTING.md#channel-description-scheme) says.

PR #2632, opened 2021-05-05
Update ar.m3u

trancer1994:
If I put a channel in the wrong country, can you get a script to automatically put it in the right one? I ask because sometimes I'm putting in so many channels that I lose track of what I've done and can't find where I've made mistakes.


freearhey (approved):

freearhey:
No, the script only checks if the link is a duplicate.

PR #2639, opened 2021-05-05
Not belonging to this list

Triloworld:
This don't belong here: 
#EXTINF:-1 tvg-id="Belsat.pl" tvg-name="Белсат" tvg-country="BY" tvg-language="Russian" tvg-logo="" group-title="",Белсат
http://f1.stream.devkom.pro:1063/ramowka/video.m3u8

Domain is not available to ensure that is Polish one


freearhey:
Channels in this playlist are sorted by the country from which they are broadcast and according to the information available to us it is Poland: https://www.lyngsat.com/tvchannels/pl/Belsat-TV.html

Triloworld:
Ok. Thanks

PR #2712, opened 2021-05-05
Update https://hiplayer.hibridcdn.net url to work with streamlink

ghost:


LaneSh4d0w:
By the way, wasn't the Rotana channels Geo-locked for a period of time ? Never had access to them with my french IP, however I could watch them using Tor...

freearhey (approved):

PR #2740, opened 2021-05-08
Remove x-tvg-url attribute

freearhey:
The `x-tvg-url` attribute is redundant because it is no longer attached to the generated playlists and its value is not used outside of it.

Also now all supported EPG codes (tvg-id) for channels can be found in [iptv-org/epg/codes.csv](https://github.com/iptv-org/epg/blob/master/codes.csv).

A list of other programs can be found here: [iptv-org/awesome-iptv#epg-sources](https://github.com/iptv-org/awesome-iptv#epg-sources).


trancer1994:
It says I deleted the branch. Have I done something wrong?

freearhey:
Don't worry, it's okay.

PR #2758, opened 2021-05-11
FR : Delete RMC

BellezaEmporium:
RMC = Radio Monte-Carlo, the "video stream" only shows an indication saying that it's live.

Not to confound with TMC, that is Télé Monte-Carlo, that is indeed a channel from the TF1 group.


freearhey (approved):

tilfizyawn:
@LaneSh4d0w I think radio streams should still be added (as long as they show video). Sometimes you want to listen to a radio station using your TV speakers.

LaneSh4d0w:
> 
> 
> @LaneSh4d0w I think radio streams should still be added (as long as they show video even if it's static). Sometimes you want to listen to a radio station using your TV speakers.

Well at some extent yes, but I think it was said not to add radio streams, so tbh this one was basically a radio stream with a plain image.

PR #2764, opened 2021-05-12
Update fr.m3u

ghost:


BellezaEmporium (changes requested):
While I would agree with this, I'm more in the idea that it would be in the XXX section, due to the mild showing of woman in soft lingerie and in sexy poses.

freearhey (approved):

PR #2766, opened 2021-05-12
Add Global Fashion Channel

ghost:


BellezaEmporium (approved):

LaneSh4d0w:
Seems to work for me, however I get an issue when VLC tries to upscale from the adaptive playlist, it just crashes on my side... Link is working though, so I cannot complain with the fact that it got it place in the playlist.

BellezaEmporium (changes requested):
Wait, something got through me. Global Fashion Channel is operated by GFC Omnimedia, an American company from Las Vegas, NV, so it's place is not in the French section, but in the US playlist.

tilfizyawn:
Thanks @LaneSh4d0w . I cannot edit us.m3u from the web because it's too large. Can you please add this channel if you can? Thanks

PR #2795, opened 2021-05-17
Add Some Locals - More to Come!

MapGuy11:
All of the streams will get a 400 error in a browser but they work great on everything else.


Sphinxroot (commented):
the main problem is , they have start dated timed stream and will change every day for the new live ...  another form of tokenized stream .. and if you remove ?start_time= section it wont play

exemple of today : ?start_time=2021-05-17T18%3A45%3A00%2B00%3A00&end_time=2021-05-17T21%3A30%3A00%2B00%3A00&env=prd&uid=12052

and it seem that we cant have the proper Referer to trick the stream , the referer is  the same for every channel ,the player seem to have a unique container for it , and not individual for each channel

MapGuy11:
Ok sorry I may have put the streams down wrong. The start date doesn't matter as it will go the latest position. All end dates can be 2023 instead of 2021 I think 2024 is too far. Then it should work beautifully. They don't token everyday they just token once for the request and index.m3u8 doesn't work without a request to the server but each resolution does. I hope this works for you.

LaneSh4d0w:
Hmm... maybe could be solutioned if the link is made by an API or some sort.
Do you have an idea from where you got those links ?

MapGuy11:
Yes, it is. It is here: https://www.localbtv.com/. Like Locast without tokens that expire.

freearhey (approved):

PR #2815, opened 2021-05-19
Obsolete: tvg-id="TeleV.ca" in ca.m3u

m-p-3:
V-Tele was [rebranded in 2020](https://en.wikipedia.org/wiki/Noovo) and the stream URL is no longer functional.

The current live stream is only available on the [the website](https://www.noovo.ca/en-direct) and is DRM-protected with Widevine.


freearhey (approved):

freearhey:
@m-p-3 Thank you for your help!

PR #2866, opened 2021-05-24
FR : Match tvg-id -> iptv-org/epg

BellezaEmporium:


LaneSh4d0w:
This is a touchy one, because names might differ depending on which website grabs the info. I grabbed the tvg-ids from our French website programme-tv.net.

PR #2955, opened 2021-05-29
Bunch of Sub Channels and Locals!

MapGuy11:


themak1985:
I added language and group titles - wondering if we should have a standardized treatment of Locals that are actually national like Grit TV, Cozy TV, PBS World, etc. 

Maybe instead of `KFMB-DT3`, it would be `Grit TV (KFMB-DT3)`

and for locals, that are actually local:
Instead of `7 NBC San Diego (KNSD-DT1)` would be `NBC San Diego (KNSD-DT1)`
Is the channel number important? Some NBC will but the number before (7 NBC) some will put it after (NBC 4). 

This is largely a US issue, but it could apply to other countries too.

themak1985 (approved):

MapGuy11:
Thanks!

PR #2971, opened 2021-05-31
PE : Modify link for Latina

BellezaEmporium:
Reason : wrong link + redirects to a 42 seconds video saying that content isn't available outside of the Peruvian territory


LaneSh4d0w:
+ Asking @sguinetti to confirm that link does not work

sguinetti:
New tv online website: https://www.latina.pe/tvenvivo

LaneSh4d0w:
Works for me. It's using MDSTRM, that does the link with the actual source. Changing link.

LaneSh4d0w:
@sguinetti Verify if this link works.

sguinetti:
Wait if this link will be preserved for hours, because the web may use token.

LaneSh4d0w:
> 
> 
> Wait if this link will be preserved for hours, because the web may use token.

This link was tokened, i took them off and it still works.

freearhey (approved):

PR #2983, opened 2021-06-01
Update pe.m3u

sguinetti:
This PR adds 4 channels more, but the periscope streaming is from https://www.pscp.tv/somostv but needs to be improved because it transmits in fragments after 24 hours.


sguinetti:
Closed for conflict of merge. Open in #2984

PR #2994, opened 2021-06-01
Update pe.m3u

sguinetti:
This pull request adds Best Cable Sports again with [Not 24/7] tag


MapGuy11 (approved):
Link Works. Thanks for adding the tag.

PR #2998, opened 2021-06-01
Fix BBC Arabic/Persian

ghost:


BellezaEmporium (approved):

LaneSh4d0w:
As always, thank you !

PR #3028, opened 2021-06-05
KR : Remove links

BellezaEmporium:
Reason : Either they're not available, either they do not answer anymore.


LaneSh4d0w:
Was made before.

PR #3038, opened 2021-06-06
update telugu channels

MKChary:


Dum4G:
You're about to add Widevine protected channels that won't work without a license key and even then will require a player that support decryption. There's no way we can have those sources in the playlist

MKChary:
> You're about to add Widevine protected channels that won't work without a license key and even then will require a player that support decryption. There's no way we can have those sources in the playlist

okay ..I will remove channels

PR #3057, opened 2021-06-07
Update clean action

freearhey:
This update:

- adds `channels/unsorted.m3u` to the weekly check list
- fixes the broken link counter
- also the script will remove all links from the playlists that respond with HTTP code 500 (Internal Server Error),  501 (Not Implemented) and 410 (Gone)

### DIFF
--- a/.github/workflows/clean.yml
+++ b/.github/workflows/clean.yml
@@ -10,7 +10,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        country: [
+        country:
+          [
             ad,
             ae,
             af,
@@ -166,8 +167,8 @@ jobs:
             vn,
             xk,
             ye,
-            zm
-            # unsorted
+            zm,
+            unsorted
           ]
     steps:
       - name: Checkout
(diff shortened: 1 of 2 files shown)


Dum4G:
Errors like 502, 503 and 504 are usually temporal, at this state PR will cause more false-positives than it should. You can add support for a rare case of 410 gone tho

freearhey:
@Dum4G Thank you for pointing that out, I've already updated the script.

BellezaEmporium (approved):

LaneSh4d0w:
You could look after the link and see if the answering result is either a m3u8 (vnd.apple.mpegurl), either an octet-stream (video/mp2t or application/octet-stream) or ts. This might reduce the number of false positives (200 and there's an url page instead)

Dum4G:
We also have some mpd streams and a few that won't return anything without a proper EXTVLC option. To reduce the number of false positives a script should repeat it's test against found links 8 hours after initial test and counter it's previous commit in the same pull request.

PR #3071, opened 2021-06-07
CU : Add channels

BellezaEmporium:
Astra streams, controlled by the ICRT.


channels/cu.m3u
```diff
@@ -1,3 +1,13 @@
 #EXTM3U
-#EXTINF:-1 tvg-id="CubaVision.cu" tvg-name="CubaVision" tvg-country="CU" tvg-language="" tvg-logo="" group-title="",CubaVision
-http://190.122.96.187:8888/http/010
```
freearhey, line 3: I'm sorry, but I don't understand why you deleted this link?

LaneSh4d0w: Not working.

freearhey: Well, it seems to be working now.
image

LaneSh4d0w: Reintegrating it, changing its name since it's the International version.

freearhey (approved):

PR #3082, opened 2021-06-08
Update unsorted.m3u

BellezaEmporium:
I don't know where to put it.


Dum4G:
https://en.m.wikipedia.org/wiki/Playboy_TV
https://www.lyngsat.com/tvchannels/us/Playboy-TV-Europe.html

PR #3097, opened 2021-06-09
Add more channels

freearhey:
- added new links from https://github.com/hginwa/md
- removed duplicate


LaneSh4d0w:
Majority of those are already in our playlists, i'll check it out after I'm done with some stuff and creating a program that parses Dum's Lyngsat text file.

freearhey:
I already ran a script looking for duplicates. Channels can be duplicated, but the links must be different from those that are already in the playlist.

LaneSh4d0w:
> 
> 
> I already ran a script looking for duplicates. Channels can be duplicated, but the links must be different from those that are already in the playlist.

Yes, was talking about the channels, not the links themselves. Of course we'll need to check the links and see if they work.

BellezaEmporium (approved):

PR #3121, opened 2021-06-11
QA: Al Jazeera cleanup, round 2

Dum4G:
Closes https://github.com/iptv-org/iptv/discussions/3118 if accepted


BellezaEmporium (commented):
Why deleting them ? They work.

Dum4G:
> Why deleting them ? They work.

Have you got the point of my discussion? We have an adaptive link that will choose the quality according to viewers bandwith. Everything else leads to the same broadcast but has fixed quality. Essentially, we can consider them as duplicates

LaneSh4d0w:
> 
> 
> > Why deleting them ? They work.
> 
> Have you got the point of my discussion? We have an adaptive link that will choose the quality according to viewers bandwith. Everything else leads to the same broadcast but has fixed quality. Essentially, we can consider them as duplicates

Thought about that after asking my question; wanting to make sure this PR was made for that.

BellezaEmporium (approved):

PR #3166, opened 2021-06-13
Removing Links That Broadcaster Doesn't Want Up.

MapGuy11:


MapGuy11:
Going Forward I am removing cities that the broadcaster elected not to have up. These cities include New York, Philadelphia, and any Eastern city. So if you see any links from app.mybtv.com with Eastern USA cities the broadcaster would not like them available to other means then their own apps.

BellezaEmporium (approved):

PR #3203, opened 2021-06-16
Update id.m3u (Replace broken Vidio links)

MinePlayersPE:
Changed vidio.com links to be tied to streamlink (since they disabled non-tokenized links)
and added live.unregister.xyz RCTIplus domain (as the only stream mirror from the site that doesn't require a Referer header)


BellezaEmporium (commented):
Excellent PR. Thanks for your changes. However, since Vidio seems to have locked certain streams, I would like to ask you to verify them once again and see if they work on your side. If they do, they're geo-locked. We would need to add the [Geo-locked] tag next to the name of the channel.

MinePlayersPE:
> I would like to ask you to verify them once again and see if they work on your side.

That's weird, now neither the old nor the new links work either. I guess they noticed? Closing because the "official" endpoint is tokenized (which is really unfortunate, since it's really the only complete set)
P.S. Streamlink should've worked, but https://linkastream.co/headless?url=https://www.vidio.com/live/204-sctv doesn't seem to work for me despite the supported sites list

LaneSh4d0w:
https://etslive-2-vidio-com.akamaized.net/exp=1623846487~acl=/vp9/204_stream.mpd!/vp9/204/*~hmac=4abd57e023456f839feef1cbc4ee9dafc75bdf73d9288865baada34c8cbc6478/vp9/204_stream.mpd

That's the kind of thing it does... had tried generating the links but those never worked.

MinePlayersPE:
Opening since it seems like [your implementation](https://query-streamlink.lanesh4d0w.repl.co/iptv-query?streaming-ip=https://www.vidio.com/live/204-sctv) works, still can't figure out why others don't
P.S. https://github.com/yt-dlp/yt-dlp/blob/7e87e27c52e9c9f0412ab360e560255533d35dce/yt_dlp/extractor/vidio.py#L228 This should give an idea of how to get the link via the API

LaneSh4d0w:
I see that my implementation works for Vidio, however I can't seem to see why linkastream wouldn't work. AFAIK we use the same solution.

BellezaEmporium (approved):
I don't see any issues on this one. All good.

PR #3232, opened 2021-06-18
US: TVS fixes

Dum4G:
From https://github.com/iptv-org/iptv/pull/3151/ with some more changes

PR #3233, opened 2021-06-18
US: TVS fixes

Dum4G:
Pulling into proper branch this time, haha. See https://github.com/iptv-org/iptv/pull/3232


LaneSh4d0w:
As said before, you can take off the "tracksv1a1" from the URLs, those should normally redirect to other URLs with tracks.

BellezaEmporium (approved):

PR #3264, opened 2021-06-20
EG : Add Al Ahly TV

BellezaEmporium:
https://en.wikipedia.org/wiki/Al_Ahly_TV


Dum4G:
Single frame and immediate kick. Is it geo-blocked?

LaneSh4d0w:
A bit of a difficulty to start but a few seconds in and works : https://streamtest.in/logs/http-149-100-19-226-8001-play-a0r6

Dum4G (approved):

PR #3307, opened 2021-06-26
CO : Add some Claro channels + RCN Novelas

BellezaEmporium:
Astra streams

### DIFF
--- a/channels/co.m3u
+++ b/channels/co.m3u
@@ -19,10 +19,16 @@ http://186.101.216.211:8000/play/a08z
 https://mdstrm.com/live-stream-playlist/574463697b9817cf0886fc17.m3u8
 #EXTINF:-1 tvg-id="CanalClaro.co" tvg-name="Canal Claro" tvg-country="LATAM" tvg-language="Spanish" tvg-logo="https://upload.wikimedia.org/wikipedia/commons/8/85/CaracolTelevisi%C3%B3n2019.png" group-title="",Canal Claro (Colombia)
 http://186.101.216.211:8000/play/a08e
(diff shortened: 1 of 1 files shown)


channels/co.m3u
```diff
@@ -59,10 +63,6 @@ https://telecaribe-hd.gcdn.anvato.net/hls/live/telecaribehd/2196k/index.m3u8
 https://telecaribe-hd.gcdn.anvato.net/hls/live/telecaribehd/3296k/index.m3u8
 #EXTINF:-1 tvg-id="Telecaribe.co" tvg-name="Telecaribe" tvg-country="CO" tvg-language="Spanish" tvg-logo="https://i.imgur.com/i5wyEor.png" group-title="",Telecaribe
 https://telecaribe-hd.gcdn.anvato.net/hls/live/telecaribehd/496k/index.m3u8
-#EXTINF:-1 tvg-id="TeleCaribeDeportes2.co" tvg-name="TeleCaribe Deportes 2" tvg-country="CO" tvg-language="" tvg-logo="" group-title="",TeleCaribe Deportes 2
```
Dum4G, line 62: I don't see a reason to remove this channel. It's not like the others

Dum4G (changes requested):

channels/co.m3u
LaneSh4d0w: It gets to a while back. Reason for this is that it shows the same programme over and over again. (just try by looking at the Channel twice, and you'll see the same programme all the time.

Dum4G (approved):

PR #3396, opened 2021-07-02
update tvg-language

Tribalize:
Updating blamk tvg-language, tvg-logo, group-title


channels/us.m3u
```diff
@@ -1,4 +1,4 @@
-#EXTM3U
+ #EXTM3U
```
Dum4G: Fixing that string should do the trick

Dum4G (changes requested):

channels/us.m3u
Tribalize: fat fingered it i guess. updated

Dum4G (approved):

Issue #3462, opened 2021-07-03
Add: Africanews


saidma:
**Channel Name:** Africanews (french) 
**Country:** International 
**Language:** french
**Category:** news
**External Link (optional):** https://fr.africanews.com/live/
https://youtu.be/b6R9-7KZ8YM

**Channel Name:** Africanews (english) 
**Country:** International 
**Language:** english
**Category:** news
**External Link (optional):** https://africanews.com/live/
https://youtu.be/NQjabLGdP5g

PR #3464, opened 2021-07-03
CG: AfricaNews

Dum4G:
Closes https://github.com/iptv-org/iptv/issues/3462


BellezaEmporium (approved):

saidma:
> Closes #3462

This channel has nothing to do with the Congo.  It only has the headquarters.  It is a Euronews channel dedicated to current affairs in Africa

LaneSh4d0w:
> 
> 
> > Closes #3462
> 
> This channel has nothing to do with the Congo. It only has the headquarters. It is a Euronews channel dedicated to current affairs in Africa

We classify the channels by their origin. Even if EuroNews possesses AfricaNews, it's HQ are in Congo, and if I remember well, the stream origin is also from this country : https://en.wikipedia.org/wiki/Africanews

Issue #3471, opened 2021-07-04
Add: Fox 2 (KTVI)


jkemp814:
**Channel Name:** Fox 2 (KTVI)
**Country:** USA
**Language:** English
**External Link (optional):** https://dcs-live.apis.anvato.net/server/play/xADY1WZSZodNcYGj/manifest.m3u8?encp=W6gAZ2M-FeKAx30p0O8SIg:vC_J05lAIy7HtzMpjJG2Qh9eqWmUM8WqeCLvju7tYtIeqWr1kgnqx5S2ztKw_AiJbwtb8I2uZErVP4kevuclowcEzrui3gjhZ8ZOyDcES5nuAtlutZMPTeY15dMIXXpX9Mbk3PpQkAJ7vr8obUFZCmoaJEh21IGm6OvQzmcQ_4S7-Yu-LDtSquNpQGBN10Upc87LXsLFBFOdbp7iCQBPisNrfdYtKciagnSlMZGMtnL7KSkTq_Hea5BdAFazMftB0GyMrPUD1yaBvTIo-NZ1DWbWEEHLtHwRv1O6kQMjKpzl1tep6eGVGlg0EJlXAk6NDx8Wt0qFPkSQGEDLsWAathVSbQ-w3ATzWg-EnR3pwkmx1_dOdt9T7SdUenJO62pe8GB4SjltRTEc9ugSlzx3dkMgANAUDleIwGHKHq5PicpvpvlY29VyGpFS0QS3aBShPPv9_HZxNq6BWK7IcST8BmMFf_Ko8LKxyUJoV3eRETTlCTvkswJrtSqsdkseREyF6SFm4nr6BbV2uUc9X0AN-T64plTDzeda6gBvlrYgDG6nwVG4E2rwm7a3yKdD8wSHDCxVcGDpJrafWSSkE0xXqvtCCEPKLJDAD0YPRYLlC4HW5Q_mDO7e14WSEhZWKFUCiElTScotlOt5xlZqFXvaQt_Z1Q9c1QURgFSbD8gEcjsldM_b_w2lG8HCoT_Wf6fKPZ_hHCgHaI7MxKGjBc_cE-4b7LkX-pNwoHNOfbrnIvI&anvtrid=w5741a7f7086a964192e9925a032b92b&anvauth=tb=0~te=1625369551~sgn=6e8a812adcc33126b33f92d55b52a03ce144720152b3a4fa690eda084e85a987&t=1625369461&_vpng=0
[...]

PR #3474, opened 2021-07-04
Template : Add additional disclaimer

BellezaEmporium:
It would be actually useful to know when we need to make additional workarounds to  have the playlist link + the website where it originates, so adding a disclaimer on the template directly.

### DIFF
--- a/.github/ISSUE_TEMPLATE/------channel-request.md
+++ b/.github/ISSUE_TEMPLATE/------channel-request.md
@@ -15,6 +15,6 @@ efficiently process your request -->
 **Channel Name:** xxx
 **Country:** xxx
 **Language:** xxx
-**Website link (necessary for any web player sources):** xxx
-**External Link (optional):** xxx
+**Broadcast page (required for any web player sources):** xxx
+**Stream Link (optional):** xxx
 **Notes (optional):** xxx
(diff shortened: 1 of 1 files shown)


freearhey:
Could you give some examples of such requests?

LaneSh4d0w:
#3471 as an example would need to get a fresh Anvato token everytime.

Count with this all "Referrer-blocked" links + "livestream.com" streams that we made working by actually grabbing a fresh link from the stream_info JSON.

Dum4G:
Take it easier

LaneSh4d0w:
I first thought about actually adding this, but I didn't know if it was useful, because we could actually put it onto Notes

Dum4G:
I'm also proposing to note this info in tvg-url parameter, this way:

It would be easier to determine origin country and language in questionable situations
It may help to determine if the source has stopped broadcasting or requires a replacement link

jkemp814:
I edited the request at #3471 should I swap the mediaManifest data with the data in External Link (optional):?

LaneSh4d0w:
> 
> 
> I edited the ADD request at #3471 should I swap the mediaManifest data with the data in External Link (optional):?

Since this PR hasn't been approved yet, you're good to keep your issue as is.

Thanks by the way for completing it with more data. It'll give us a better look on this stream.

freearhey (approved):

PR #3504, opened 2021-07-10
Added missing channel logo's

jkemp814:
Few more logo's


jkemp814:
CBN News and CBN USA has the logo in the group title. How can I correct that?

Line number: 663, 667

PR #3539, opened 2021-07-15
Fast patch - Change group title

BellezaEmporium:
Group title shouldn't be "Public Broadcasting" but "General"


MapGuy11 (approved):

jkemp814:
There seem to be an error at line 6595 in the name. Geo block is there twice.

PBS [Geo-blocked]rgia

PR #3594, opened 2021-07-20
Add Pluto TV En Español

adamford21:
Pluto TV En Español


LaneSh4d0w:
1) Not respecting the Contribution guide
2) The channels aren't in the correct playlist (we won't create a playlist specially for Pluto TV)
3) I do think we have most of the channels you're adding.

PR #3614, opened 2021-07-21
Channel edits - Filling in missing data

jkemp814:
Work continues.


jkemp814:
I have a question: 

I have channels that are in the us.m3u playlist, but they don't show up in VLC and Kodi. In some cases I have more channels in the us.m3u playlist but only a couple show in VLC. 

Black News Channel, for example. I have over six in the playlist but only two in VLC or Kodi. 

I am in sync with my remote branch and the upstream branch. What would cause this?

The new 'VN' and Brazil channels are behaving like that also.

Dum4G:
Generated playlists (index country/language/group ones) contain the first link for a channel only and it's determined simply in alphabetical order, assuming all of the sources are equally cool for everyone and will be used as a backup plan when the first link dies. If you see duplicated channels in generated playlists that means they're having different tags or scripts doesn't act perfectly (for ex. a source with http referer or user-agent is considered a different channel)

jkemp814:
Okay. so that's normal? Maybe because I have been syncing all the tags with channels, that's what cause them to not show up in those two apps? I remember seeing all ten channels in VLC awhile back

Dum4G:
Yeah, you're "normalising" the us playlist to the point where you won't see duplicate channels unless there are reasons that you can't fix

jkemp814:
I actually like that. I was concerned that I wasn't in sync with the master branch and would cause some kind of problem going forward.

Dum4G:
Don't forget that the master branch is the master branch and changes immediately as we merge into it, but gh-pages are syncing at fixed time which is 0:00 GMT (or if someone with access will manually trigger it)

jkemp814:
I was wondering about that. So as soon as a merge is performed, I should see the changes in the playlist? I was always waiting until the update-playlist bot ran.

jkemp814:
There was a white space tvg-language="English " do the checks pass this if it's with in the quotes?

channels/us.m3u
```diff
@@ -530,13 +530,13 @@ https://bloomberg-bloombergtv-1-es.samsung.wurl.com/manifest/playlist.m3u8
 https://liveprodusphoenixeast.akamaized.net/USPhx-HD/Channel-TX-USPhx-AWS-virginia-1/Source-USPhx-16k-1-s6lk2-BP-07-02-81ykIWnsMsg_live.m3u8
 #EXTINF:-1 tvg-id="BloombergUSDesktop.us" tvg-name="Bloomberg US Desktop" tvg-country="US" tvg-language="English" tvg-logo="http://cdn-profiles.tunein.com/s47135/images/logog.png" group-title="Business",Bloomberg US Desktop
 https://liveprodusphoenixeast.global.ssl.fastly.net/USPhx-HD/Channel-TX-USPhx-AWS-virginia-1/Source-USPhx-16k-1-s6lk2-BP-07-03-0Yn1cQZHOtP_live.m3u8
-#EXTINF:-1 tvg-id="BloombergMeiGuo.us" tvg-name="Bloomberg(美国)" tvg-country="US" tvg-language="English" tvg-logo="http://cdn-profiles.tunein.com/s47135/images/logog.png" group-title="Business",Bloomberg(美国)
+#EXTINF:-1 tvg-id="BloombergMeiGuo.us" tvg-name="Bloomberg (美国)" tvg-country="US" tvg-language="English" tvg-logo="http://cdn-profiles.tunein.com/s47135/images/logog.png" group-title="Business",Bloomberg (美国)
```
Dum4G, line 533: Not chinese version or source, chinese part should be removed entirely

Dum4G (changes requested):

channels/us.m3u
```diff
@@ -530,13 +530,13 @@ https://bloomberg-bloombergtv-1-es.samsung.wurl.com/manifest/playlist.m3u8
 https://liveprodusphoenixeast.akamaized.net/USPhx-HD/Channel-TX-USPhx-AWS-virginia-1/Source-USPhx-16k-1-s6lk2-BP-07-02-81ykIWnsMsg_live.m3u8
 #EXTINF:-1 tvg-id="BloombergUSDesktop.us" tvg-name="Bloomberg US Desktop" tvg-country="US" tvg-language="English" tvg-logo="http://cdn-profiles.tunein.com/s47135/images/logog.png" group-title="Business",Bloomberg US Desktop
```
Dum4G, line 531: I don't think it's really called US Desktop

Dum4G (changes requested):

channels/us.m3u
jkemp814: Okay, I have seen that in a few others

channels/us.m3u
jkemp814: I have seen that last week, I think they were trying to say the website version, will go take a look.

jkemp814: That version does look different, seems like a desktop app

jkemp814: Maybe we should name it Bloomberg Anywhere

LaneSh4d0w: Link above the US Desktop is the same.

jkemp814: Line 530 and 532 are not the same, not an exact match on my local branch

jkemp814: Are you saying it might be Bloomberg TV+ I checked there are many of those and they're broadcasting the same content. I will rename it to Bloomberg TV+

Dum4G (approved):

PR #3620, opened 2021-07-21
Update and rename channels/jp.m3u to 頻道/jp.m3u

tonyt89:


Dum4G (approved):

Dum4G:
I can approve channels you did in file but not folder rename

PR #3639, opened 2021-07-22
LATAM + ES : Add even more channels

BellezaEmporium:


LaneSh4d0w:
Quick but interesting one for Spanish people.

(will have some DAZN channels coming soon)

Dum4G (approved):

PR #3650, opened 2021-07-23
Update jp.m3u

vbHeiko:


BellezaEmporium (commented):
Already available on the US playlist as Disney Channel Japan.

vbHeiko:
I think Disney Channel Japan should be in the JP playlist because many users don't know it's in the US playlist.

LaneSh4d0w:
We classify their channels by their origin, and since Disney Channel is American-based, it's on the US playlist. However, it's weird that you do not see Disney Channel Japan on the JP playlist. Let me take a look.

vbHeiko:
A few days ago someone asked for jp.m3u disney channel that's why I spend an hour decompiling an Android app. I found Disney Channel in the app source code (lool) and made a pull request that was rejected. I think people downloading the JP channel list are more interested in channels in Japanese (Japanese language) than they care about the channel's origin. So I think it would be cool to get the Disney channel on the JP channel list.

PR #3658, opened 2021-07-24
Pluto TV - Second round of sanitation

jkemp814:
Work continues.


Dum4G:
That's exactly what I thought to prevent from happening. If a channel doesn't show any content exclusive to PlutoTV or any other platform I'd leave it as just another backup source with identical tags to other links IMO

LaneSh4d0w:
To be honest, i'm not even sure the channels Pluto host are "real" channels. Here in France, it's a mishmash of movies and TV shows being put onto a channel.

Might be the same for the US and all the other parts.

jkemp814:
Those hand full of channels I changed already were coming from pluto.tv and had Pluto TV at the end of the description. Also, they fix right in next to the other channels that were already down in the Pluto list. There are no more to rename or move. It was 36 of them that were different from the other Pluto TV channels. I think it was right to move them where they need to be. They are Pluto TV channels. Renaming them was just making the list consistent.

Dum4G (approved):

PR #3706, opened 2021-07-26
Changing names of news stations: Moving call sign to end

jkemp814:


jkemp814:
Comma in name cause lint check to think there is a white space

Dum4G (approved):

PR #3725, opened 2021-07-27
Update ba.m3u, hr.m3u, rs.m3u, si.m3u

muzena:

PR #3739, opened 2021-07-28
Ex-Yu: cleanup and channel additions

Dum4G:
Following suggestions from https://github.com/iptv-org/iptv/pull/3725
Thanks @muzena. SPTV won't be included into the list since it's geo-blocked and we don't have Croatian Streamlink proxy for DM links


channels/ba.m3u
```diff
@@ -3,12 +3,8 @@
 http://proxy.bihnet.net:88/live/b1tv.m3u8
 #EXTINF:-1 tvg-id="BHRT.ba" tvg-name="BHRT" tvg-country="BA" tvg-language="Bosnian" tvg-logo="https://i.imgur.com/F8jTl1o.png" group-title="",BHRT (720p)
 https://bhrtstream.bhtelecom.ba/bhrtportal.m3u8
```
LaneSh4d0w, line 5: Those links sound Geoblocked to me, I never could launch them.

BellezaEmporium (approved):

PR #3751, opened 2021-07-29
Update kp.m3u

jinniryu:
Add another m3u8 source


jkemp814:
Stream will not play. May be [Geo-blocked]

Sphinxroot (commented):
403 Forbidden  could be geo

Dum4G (approved):

PR #3813, opened 2021-07-30
Removed comma in channel description

jkemp814:
Apparently when there is a comma between city and state in the channel description, it will only show what's defined after the comma. Even though it doesn't throw an error with the lint checker. 

,NBC 2 Buffalo, NY (WGRZ) Shows up as NY (WGRZ)


jkemp814:
Should the Playboy TV Latin America be showing up in the us.m3u playlist?

Line 3166

Dum4G (approved):

PR #3814, opened 2021-07-30
Add PBO Digital

sguinetti:
This PR adds PBO Digital peruvian channel with main programation on the morning and afternoon.


Dum4G:
That link is temporarily down. Please use https://linkastream.co/headless?url= instead

sguinetti:
Ok. Thanks.

Dum4G (approved):

PR #3876, opened 2021-08-03
Matching Channels with EPG

jkemp814:
Syncing channels with EPG list.


channels/us.m3u
```diff
@@ -25,21 +25,21 @@ https://30a-tv.com/feeds/xodglobal/30atv.m3u8
...
 https://moiptvhls-i.akamaihd.net/hls/live/652312/secure/master.m3u8
 #EXTINF:-1 tvg-id="3ABNKids.us" tvg-name="3ABN Kids" tvg-country="US" tvg-language="English" tvg-logo="https://i.imgur.com/eOvzf3d.png" group-title="Religious",3ABN Kids
 https://moiptvhls-i.akamaihd.net/hls/live/652318/secure/master.m3u8
-#EXTINF:-1 tvg-id="3ABNLatino.us" tvg-name="3ABN Latino" tvg-country="US" tvg-language="Spanish" tvg-logo="https://i.imgur.com/4ydVVIA.png" group-title="Religious",3ABN Latino
+#EXTINF:-1 tvg-id="K17JID4.us" tvg-name="3ABN Latino" tvg-country="US" tvg-language="Spanish" tvg-logo="https://i.imgur.com/4ydVVIA.png" group-title="Religious",3ABN Latino
 http://uni5rtmp.tulix.tv:1935/bettervida/bettervida/playlist.m3u8
-#EXTINF:-1 tvg-id="3ABNLatino.us" tvg-name="3ABN Latino" tvg-country="US" tvg-language="Spanish" tvg-logo="https://i.imgur.com/4ydVVIA.png" group-title="Religious",3ABN Latino
+#EXTINF:-1 tvg-id="K18JLD2.us" tvg-name="3ABN Latino" tvg-country="US" tvg-language="Spanish" tvg-logo="https://i.imgur.com/4ydVVIA.png" group-title="Religious",3ABN Latino
```
LaneSh4d0w, line 38: Same programmation than the one above, but different call signs ?

jkemp814:
Yes. There are about eight of those. Used different call signs to see which channel it might be base on time zones and content being played.

LaneSh4d0w:
I see.

Even https://en.wikipedia.org/wiki/List_of_Three_Angels_Broadcasting_Network_affiliates that contains a lot of cities and their call signs is pretty confusing to read, when we're searching at the same time x)

channels/us.m3u
jkemp814: When streaming, some of them were showing different content. So, I'm trying to match them to the EPG. All have different call signs.

Dum4G:
Am I approving it?

jkemp814:
Yes I have to leave, so I will open a new one, later.

Dum4G (approved):

PR #3879, opened 2021-08-03
UA : Fastpatch

BellezaEmporium:


LaneSh4d0w:
If you make them communicate on HTTPS using port 8081, it says that the SSL protocol answer's too long, hence the https and 8083 on this particular server.

Dum4G (approved):

PR #3883, opened 2021-08-03
CY: New channels

Dum4G:


channels/gr.m3u
```diff
@@ -241,3 +241,7 @@ http://5db313b643fd8.streamlock.net:1935/ZerounoTV/ZerounoTV/playlist.m3u8
 https://zouglalive-lh.akamaihd.net/i/zouglalive_1@340792/master.m3u8
 #EXTINF:-1 tvg-id="Krete TV GR" tvg-name="Krete TV GR" tvg-country="GR" tvg-language="Modern Greek (1453-)" tvg-logo="https://i.imgur.com/Ldt07Lh.png" group-title="",ΚρήτηTV (720p)
 http://live.cretetv.gr:1935/cretetv/myStream/f1tv.m3u8
+#EXTINF:-1 tvg-id="ANT1CYPRUS.GR" tvg-name="ANT1 Cyprus" tvg-country="CY" tvg-language="Modern Greek (1453-)" tvg-logo="https://www.worldtvpc.com/onlinevideosites/img/ant1.jpg" group-title="",ANT1 Cyprus
+https://www.ant1.com.cy/ajax.aspx?m=Atcom.Sites.Ant1iwo.Modules.TokenGenerator&videoURL=https://l2.cloudskep.com/antl2/abr/playlist.m3u8
```
LaneSh4d0w, line 245: I don't think that one will work, because it's a text that says where the link is. Few seconds.

Dum4G: Works in VLC for me

LaneSh4d0w: Interesting. Wouldn't predict that would work lol. Well then, here's an alternative one in case of (just redirects in the URL inside the link : https://iptv-all.lanesh4d0w.codes/cyprus/ant1

pouyiouk: RIK S and RIK SAT are the same channels. Duplicate entries.

pouyiouk: ANT1cy and VOULIcy should be under cy.m3u

Dum4G: We're referencing lyngsat to determine an origin country. According to them those channels have headquarters in Greece. Those will be present in Cyprus playlist tho

pouyiouk: Oh okay thanks for letting me know, wasn't aware of that. There is also an outdated link in lyngsat for VOULI cy that is not working. Latest link for ANT1cy is not playing under VLC. Line 226/227 in gr.m3u

LaneSh4d0w: They're changing servers I suppose. Let me see.

pouyiouk (approved):

channels/gr.m3u
LaneSh4d0w: It seems it's the same server. I haven't made a mistake when I've made the workaround either.

Maybe there's a specific user-agent to use, since it works on computer. @Dum4G knows this aspect more than I do.

LaneSh4d0w: @Dum4G Try my link with referer positioned in https://www.ant1.com.cy/

Dum4G: Didn't work

LaneSh4d0w: Interesting... there's something in this link that blocks then.

LaneSh4d0w: Just made Alpha Cyprus, pushing it soon onto my server. Works perfectly on my watch.

Testing a plan B for Ant1.

Dum4G: OG link to a function is no longer providing working link. Maybe needs some proper orgin or has become geo-blocked

LaneSh4d0w: It's geoblocked in its root. But even I with a Cyprus IP can only watch it using the player. Pretty sure there's a sort of referrer or origin lock.

LaneSh4d0w: Something sounds weird.

image

LaneSh4d0w: Found the issue, pushing the changes.

LaneSh4d0w: Ant1 is the same link as the one i've sent above.
Alpha : https://iptv-all.lanesh4d0w.codes/cyprus/alpha-tv

LaneSh4d0w: Omega TV : https://iptv-all.lanesh4d0w.codes/cyprus/omega-tv
Vouli (in case the link in list goes down) : https://iptv-all.lanesh4d0w.codes/cyprus/vouli

LaneSh4d0w: It seems that all the Cloudskep links cease to work once I push them on my server, but works perfectly when being put on my server. Might be because they're geo-locked even at that scale... Dang it.

LaneSh4d0w:
I've added a Cypriot proxy behind Ant1, can someone verify using a Cypriot VPN ?

danielski1988 (approved):

Dum4G:
What's app with Alpha, Omega and ANT1? Is there anyone from Cyprus to tell if everything else works?

Dum4G:
@freearhey is there a reason to use Modern Greek (1453-) and not just Greek?

LaneSh4d0w:
> What's app with Alpha, Omega and ANT1? Is there anyone from Cyprus to tell if everything else works?

I've added a Cypriot proxy on Ant1 only for now, for testing purposes. Awaiting confirmation that it works.

Dum4G:
> I've added a Cypriot proxy on Ant1 only for now, for testing purposes. Awaiting confirmation that it works.

Alpha and Omega should work without it. At least I was able to see them without VPN and grab my personal link into VLC just fine

LaneSh4d0w:
451 means that content is not available due to legal reasons. Another way of saying "geo-locked".

freearhey:
@Dum4G The language names are taken directly from [iso639-3.sil.org](https://iso639-3.sil.org/code_tables/639/data) and there is simply no other option.

Dum4G:
@LaneSh4d0w does Alpha and Omega offer wmsAuthSign on per-IP basis? If so I'll make this PR ready for approval

LaneSh4d0w:
> @LaneSh4d0w does Alpha and Omega offer wmsAuthSign on per-IP basis? If so I'll make this PR ready for approval

I'll need to recheck, but on my internal testings : Alpha and Omega worked with my Cypriot VPN IP (that I can't use for now)

LaneSh4d0w:
Alpha works without VPN if contacted from here : http://stgtwo.alphatv.gr/page/live#gsc.tab=0

LaneSh4d0w:
Omega however doesn't work without a VPN in.

LaneSh4d0w:
Alpha TV should now work fine.

LaneSh4d0w:
Added a Cypriot proxy to Omega TV

LaneSh4d0w:
Well... each time I put it onto my server, it doesn't work.

Dum4G:
> Well... each time I put it onto my server, it doesn't work.

Your implementation doesn't have nimble id and I assume it's per-ip based. I can watch both from the website without VPN. ANT1 says "broadcast temporarily unavailable"

LaneSh4d0w:
> 
> 
> > Well... each time I put it onto my server, it doesn't work.
> 
> Your implementation doesn't have nimble id and I assume it's per-ip based. I can watch both from the website without VPN. ANT1 says "broadcast temporarily unavailable"

Interesting, thanks for telling me, i'll review my implementation.

Dum4G:
I think I'm having different wmsAuthSign, so it won't be bypassable

BellezaEmporium (approved):

PR #3895, opened 2021-08-04
ABC 2 - Edited call sign

jkemp814:
This one got overlooked


BellezaEmporium (approved):
We'll see later concerning the pictures. We need to see how we can turn the URL changes onto our advantage.

PR #3898, opened 2021-08-04
Fix channel description

freearhey:
Fixes incorrectly detected stream resolution.

Resolves: https://github.com/iptv-org/iptv/pull/3897#issuecomment-892350880


BellezaEmporium (approved):
I might blame the resolution detector script for this one. Some adaptive streams got "audio_only" playlists before switching to video and audio.

PR #3986, opened 2021-08-07
Channel edits - Sync logos - Rename - Change call sign - Added state

jkemp814:
Corrected error with call sign before city.


jkemp814:
Some has state initials, some didn't. Trying to keep the list consistent.

jkemp814:
I can't spell 'Change'

Dum4G (approved):

PR #4016, opened 2021-08-09
Update "clean" script

freearhey:
This update replaces the HTTP client previously used to check links with [ffmpeg](https://www.ffmpeg.org/) library. Since FFmpeg uses basically the same codecs as VLC player, Kodi and many other popular apps (https://en.wikipedia.org/wiki/Libavcodec), this change should allow us to determine broadcast status much more accurately.


Dum4G:
A website we promote in awesome-epg repo Jacal.sh recommends to use plain HTTP because quite a lot links may not play if there's an HTTPS conflict. Shall we be worried about that?

Dum4G (approved):

PR #4034, opened 2021-08-10
Update "format" script

freearhey:
After this update the `format` script will be able to :

- add `tvg-id`, `tvg-name` and `tvg-country` if not specified
- set stream resolution if not specified
- remove invalid `group-title`
- check stream status
- normalize url

Also now broken links will not be removed from playlists every week. Instead, each link will be checked daily and broken ones will be marked `[Offline]` as suggested [here](#4019).

Test results: https://github.com/iptv-org/iptv/commit/3365d6a47df82b29f9761884d02c2185c51015a0


Dum4G (approved):

Dum4G:
Needs a quick fix for a normalizer that breaks Streamlink queries

freearhey:
@Dum4G Could you give me an example of such a link?

Dum4G:
https://linkastream.co/headless?url=https://www.dailymotion.com/video/x74wje5
https://linkastream.co/headless?url=https%3A%2F%2Fwww.dailymotion.com%2Fvideo%2Fx74wje5

freearhey:
@Dum4G Sorry, but I still do not understand what exactly the problem is.

The first link does not work even without normalization:
image

And the second one works fine just like the original:
image

Dum4G:
There's no confidence that Streamlink servers will work with normalized links when it's back online, same can be said about other tokenized sources

freearhey:
I wouldn't worry about it, since encoding a query string is common practice: https://en.wikipedia.org/wiki/Percent-encoding

But for readability I think it would be really better to decode the links back.

Here's the updated version: #4043

PR #4036, opened 2021-08-10
Revert "Update "format" script"

Dum4G:
Reverts iptv-org/iptv#4034 if @freearhey won't be able to fix normalizer today. I have restored the update-format-script branch


LaneSh4d0w:
Dang, that's lots of lines being moved / changed.

Dum4G:
It brings all the scripts back to a previous state. All of the progress being made will remain in a separate branch

BellezaEmporium (approved):

PR #4043, opened 2021-08-11
Update "format" script

freearhey:
After this update the `format` script will be able to :

- add `tvg-id`, `tvg-name` and `tvg-country` if not specified
- set stream resolution if not specified
- remove invalid `group-title`
- check stream status
- normalize url

Also now broken links will not be removed from playlists every week. Instead, each link will be checked daily and broken ones will be marked `[Offline]` as suggested [here](#4019).


Dum4G:
Is it gonna remove an [Offline] tag if the source is back online?

freearhey:
Yes

Dum4G (approved):

PR #4068, opened 2021-08-14
Add ok.ru links back

ghost:
I'm not sure why https://github.com/iptv-org/iptv/pull/4056/files deleted them


BellezaEmporium (approved):

freearhey:
Links have been removed because they do not work at the moment and have an alternative in the same playlist. Here is a more detailed description of how it works: https://github.com/iptv-org/iptv/pull/4055

tilfizyawn:
Thanks @freearhey that makes sense. When linkastream.co is back they can be re-added

tilfizyawn:
@freearhey I think the delete should be done once a week with a pull request. Sometimes the links are valid but just down.

freearhey:
@tilfizyawn I think you're right, it would be better to do just that.

freearhey:
@tilfizyawn Done via https://github.com/iptv-org/iptv/pull/4116

PR #4070, opened 2021-08-14
Update al.m3u

boomski:
Added all RTSH channels (Albanian Public TV)


BellezaEmporium (changes requested):
Works for me, however i'm not sure the "AL |" at the end of the stream name is useful.

boomski:
> 
> 
> Works for me, however i'm not sure the "AL |" at the end of the stream name is useful.

i removed the "AL |" .

BellezaEmporium (approved):

PR #4074, opened 2021-08-14
Edit My Domain

MapGuy11:
In recent weeks I have been getting a lot of traffic to my redirect url. With this I am changing to Netlify and Cloudflare for a good distribution and lower latency times. If you cannot access my domain it could be your DNS server. Also this domain will have higher up time as I am not self hosting it anymore.


BellezaEmporium (approved):
Good luck dealing with the extra requests haha

MapGuy11:
> Good luck dealing with the extra requests haha

Now A bunch of people's copycats lists are going to be dead :)

PR #4178, opened 2021-08-18
Add USIL TV

sguinetti:
This PR adds other peruvian channel. I don't sure if this channel is 720p, but the web is https://usil.tv/index.php/channel/12/se%C3%B1al-en-vivo/


Dum4G:
Don't worry, it is 720p

Dum4G (approved):

PR #4227, opened 2021-08-22
Update cleanup workflow

freearhey:
To avoid deleting working links, the PR created by `iptv-bot` will have to be approved manually.

Resolves #4225


Dum4G:
Can we automate the process in a way where [Offline] tag gets replaced with [Not 24/7] when needed? I thought there's no need to do manual and tbh based checks since we have new scripts

freearhey:
Good idea, I'll do that.

PR #4242, opened 2021-08-23
Update "cleanup" workflow

freearhey:
After the update:

- the script will mark `[Offline]` links as `[Not 24/7]` if they work again
- the weekly removal of `[Offline]` links will have to be approved manually to avoid any mistakes


BellezaEmporium (approved):

Dum4G:
I still don't understand why to remove [Offline] links if we can preserve our knowledge about those channels? (Tvg-country, origin, logo, group, proper name)

freearhey:
The script only removes [Offline] links that have alternatives, so all that information stays in the playlist.

PR #4264, opened 2021-08-24
Update lb.m3u

sroykh:
LBC International LOGO was MBC ones


Dum4G:
Are you sure it wasn't glitch in a player? Both look identical to me, but old one has more appropriate size

Sphinxroot:
same logo but

have = 220x424
changed = 42x104

sorry  we will keep it that way

PR #4303, opened 2021-08-28
hta.dz is tokenized now

Dum4G:
Mark it as offline till solution comes


LaneSh4d0w:
The only way I could make them work is by generating the token specifically with a proxy that I should divulgate after.

freearhey (approved):

PR #4397, opened 2021-09-06
CO : Add some channels

BellezaEmporium:
Needs either origin or referrer to get it work.


LaneSh4d0w:
Hehe... that source is giving me some headaches.

LaneSh4d0w:
For the time being, I won't add Win Sports, due to some gateway timeout issues (pretty sure from what it's caused)

Dum4G (approved):

PR #4467, opened 2021-09-11
Create us_xumo.m3u

freearhey:
Creates a separate file for channels broadcast via XUMO.

https://www.xumo.tv/channels


Sphinxroot (approved):

jkemp814:
How are you wanting to handle channels with (XUMO) at the end of the channel description? Do you want it removed? Should it be placed at the beginning of the tvg-id tag? There are two remaining channels in the us.m3u file, should those be moved to the newly created us_xumo.m3u file? (i.e. Football Daily (XUMO), Motorvision (XUMO)?

freearhey:
@jkemp814 As I understand it, these channels are broadcast through another OTT service FuboTV.

jkemp814:
Okay, I'll wait on your queue before changing any of them.

PR #4527, opened 2021-09-13
Update il.m3u

sibonye:


Dum4G:
1. We do not replace links that work. If you want to add an alternate source just duplicate all tags of a channel and do a separate entry.
2. This is an Xstream link. Those are usually limited to low amount of viewers and a period of subscription time. It won't last long

PR #4551, opened 2021-09-14
CN : Change certain links + clean some others

BellezaEmporium:
- Moved all links that were "working" onto http://m-tvlmedia.public.bcs.ysten.com/
- The ones that couldn't be moved were deleted, waiting for a solution linked to them.
- The fact I didn't push changes directly were due to the massive number of links deleted that couldn't be moved, that's why I *deliberately* made a separate branch for the cleanup one.


Dum4G:
We should keep info about those channels just in case we'll ever find an alternative, shall we?

LaneSh4d0w:
Gonna dig a bit further on this one, see if I can manage to get the origin for the other IP or if I can get the stream (maybe they've changed its name)

LaneSh4d0w:
Found a CDN that did the same stuff than the original ysten.com links. Added them as alternatives.

Dum4G (approved):

PR #4610, opened 2021-09-18
Update format.js

freearhey:
After this update links that return a 403 HTTP code will be marked as [Offline].


taodicakhia:
Are you sure that all update links returning 403 HTTP code will be marked as [Offline]? Some links only add referrer (user-agent) are playable.

freearhey:
The script takes user-agent and referrer into account when checking links.

Dum4G (approved):

PR #4643, opened 2021-09-21
Fix Brazil EPG

allanf181:


freearhey:
Where exactly did you get the new tvg-id from?

allanf181:
https://siptv.app/codes/

Enviado do Yahoo Mail no Android 
 
  Em ter., 21 21e set. 21e 2021 às 4:57, Aleksandr ***@***.***> escreveu:   


Where exactly did you get the new tvg-id from?

freearhey (changes requested):
The thing is that we try to use our own [code list](https://iptv-org.github.io/epg/index.html) where possible to add a program to as many channels as possible.

allanf181:
I don't know if it has a standard but other iptv providers in Brazil use the same codes as https://siptv.app/codes/

allanf181:
anyway there are other things in my fix like channels with wrong name and duplicate channels

freearhey:
There is no standard and each provider can use any list of codes.

What duplicates are we talking about?

allanf181:
https://github.com/iptv-org/iptv/blob/master/channels/br.m3u#L100
https://github.com/iptv-org/iptv/blob/master/channels/br.m3u#L222

freearhey:
Now I see.

freearhey:
@allanf181 you still need to bring back the old id so we can approve the rest of the changes

allanf181:
done

freearhey:
please check again

freearhey (approved):

PR #4691, opened 2021-09-24
Update pk.m3u

Sphinxroot:
added a couple channel  removed PTV because its a PH Stream


Sphinxroot:
see PH PR  wrong country

Dum4G (approved):

PR #4700, opened 2021-09-24
Update ca.m3u

AlexandreBouffard:
- Align tvg-id with iptv-org/epg
- Update attributes
- Higher resolution stream URL

### DIFF
--- a/channels/ca.m3u
+++ b/channels/ca.m3u
@@ -139,7 +139,9 @@ http://cdn12.henico.net:8080/live/gsctv/index.m3u8
 https://ici-i.akamaihd.net/hls/live/873426/ICI-Live-Stream/master.m3u8
 #EXTINF:-1 tvg-id="IciRDI.ca" tvg-country="CA" tvg-language="French" tvg-logo="https://raw.githubusercontent.com/geonsey/Free2ViewTV/master/images/logos/ICI-RDI.png" group-title="News",Ici RDI (720p)
 https://rcavlive.akamaized.net/hls/live/704025/xcanrdi/master.m3u8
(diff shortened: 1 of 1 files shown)


channels/ca.m3u
```diff
@@ -139,8 +139,8 @@ http://cdn12.henico.net:8080/live/gsctv/index.m3u8
 https://ici-i.akamaihd.net/hls/live/873426/ICI-Live-Stream/master.m3u8
 #EXTINF:-1 tvg-id="IciRDI.ca" tvg-country="CA" tvg-language="French" tvg-logo="https://raw.githubusercontent.com/geonsey/Free2ViewTV/master/images/logos/ICI-RDI.png" group-title="News",Ici RDI (720p)
 https://rcavlive.akamaized.net/hls/live/704025/xcanrdi/master.m3u8
-#EXTINF:-1 tvg-id="ICITele.ca" tvg-country="CA" tvg-language="French" tvg-logo="https://i.imgur.com/byTTNAi.png" group-title="General",ICI Tele (720p)
-https://rcavlive.akamaized.net/hls/live/696615/xcancbft/master.m3u8
```
freearhey, line 143: This link works fine
image
and the new one only works through a VPN

So it would be better to leave both links in the playlist and mark the new one as `[Geo-blocked]`.

freearhey (changes requested):

channels/ca.m3u
AlexandreBouffard: Makes sense, could not test new link using streamtest.in (reported everything as offline). Fixed commit.

freearhey (approved):

PR #4769, opened 2021-09-27
Create us_adultiptv.m3u

freearhey:
Creates a separate file for channels broadcast via AdultIPTV.net.

https://adultiptv.net/

#4768


Dum4G (approved):

LaneSh4d0w:
Ehh... isn't AdultIPTV.net a French website ? At least they go through OVH France for their streams.

freearhey:
@LaneSh4d0w I only added this prefix because I couldn't find any connection to the other country, but if you have other data, we can change it.

LaneSh4d0w:
> @LaneSh4d0w I only added this prefix because I couldn't find any connection to the other country, but if you have other data, we can change it.

Oh okay, no problem then :)

PR #4788, opened 2021-09-28
Update us.m3u

mmkogut:
Fixed/replaced broken NBC 4 Los Angeles Link

### DIFF
--- a/channels/us.m3u
+++ b/channels/us.m3u
@@ -1165,8 +1165,8 @@ http://stream.ec.nau.edu/live/amlst:channelfour/playlist.m3u8
 https://www.livedoomovie.com/02_nbahd_720p/chunklist.m3u8
 #EXTINF:-1 tvg-id="WGRZTV.us" tvg-country="US" tvg-language="English" tvg-logo="https://i.imgur.com/m4MZTtj.jpg" group-title="Local",NBC 2 Buffalo NY (WGRZ) (1080p)
 https://livevideo01.wgrz.com/hls/live/2016286/newscasts/live.m3u8
-#EXTINF:-1 tvg-id="KNBCTV.us" tvg-country="US" tvg-language="English" tvg-logo="https://upload.wikimedia.org/wikipedia/commons/thumb/e/ed/KNBC_%282011_-_Flat%29.svg/1280px-KNBC_%282011_-_Flat%29.svg.png" group-title="Local",NBC 4 Los Angeles CA (KNBC) (1080p) [Offline]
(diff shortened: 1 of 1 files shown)


MapGuy11:
I added @LaneSh4d0w to see what he thinks I approve it though.

MapGuy11 (approved):

BellezaEmporium (approved):

mmkogut:
Thanks everyone

LaneSh4d0w:
@MapGuy11 You're the one monitoring this stream, you don't need my approval for that, as long as the information is accurate :)

MapGuy11:
> @MapGuy11 You're the one monitoring this stream, you don't need my approval for that, as long as the information is accurate :)

Yeah it is. :)

jkemp814:
A new channel should have been added, and the original newscast stream fixed, not replaced. 

There are viewers who prefer the newscast over a live full schedule channel. News is what's sought after. Not only did we replace an existing channel whether broken or not, in its place, added a channel with a whole different category.

LaneSh4d0w:
> 
> 
> A new channel should have been added, and the original newscast stream fixed, not replaced.
> 
> There are viewers who prefer the newscast over a live full schedule channel. News is what's sought after. Not only did we replace an existing channel whether broken or not, in its place, added a channel with a whole different category.

Pretty sure we should actually indicate whether the channel the end user is about to see is the Newscast or the OTA channel.
We did no indication whatsoever.

jkemp814:
I have wondered about that as I add all these newscast channels. If you look at the ABC Los Angeles channels, one uses DT1 full schedule and the other -TV is the newscast. The end user/viewer may not notice that. 

It's best to keep both instead of replacing URL's for what viewers perceive as a better channel because it's live. We lose the newscast doing it that way. Why not have both?  With name real estate an issue sometimes, adding Newscast makes the description longer. Adding (NC) may not be understood also.

LaneSh4d0w:
> 
> 
> I have wondered about that as I add all these newscast channels. If you look at the ABC Los Angeles channels, one uses DT1 full schedule and the other -TV is the newscast. The end user/viewer may not notice that.
> 
> It's best to keep both instead of replacing URL's for what viewers perceive as a better channel because it's live. We lose the newscast doing it that way. Why not have both? With name real estate an issue sometimes, adding Newscast makes the description longer. Adding (NC) may not be understood also.

[OTA] [News] sounds OK to me.

mmkogut:
A user's perspective, I find it currently easy to tell between the newscast channels and OTA when they have an 'N' placed after the station or network.
i.e. "CBSN Boston" "ABC-N Los Angeles" compared to "CBS 8 San Diego" or "NBC 4 Los Angeles" which are the OTA's.

Hope that might help.

PR #4795, opened 2021-09-29
CO : Add Zoom

BellezaEmporium:


LaneSh4d0w:
Having a really weird issue with connection going in successfully, it gets the chunklist, however it locks on 403 just after this when getting the .ts files...

PR #4839, opened 2021-10-02
Update gl.m3u

Sphinxroot:


freearhey (approved):

channels/gl.m3u
```diff
@@ -1,5 +1,5 @@
 #EXTM3U
-#EXTINF:-1 tvg-id="KNR1.gl" tvg-country="GL" tvg-language="Greenlandic (Kalaallisut);Danish" tvg-logo="https://i.imgur.com/uQ50JUX.jpg" group-title="",KNR1 (1080p) [Offline]
-https://query-streamlink.herokuapp.com/iptv-query?streaming-ip=https://www.youtube.com/channel/UCIjOAtv53RHerlfl98fZmcQ/live
+#EXTINF:-1 tvg-id="KNR1.gl" tvg-country="GL" tvg-language="Greenlandic (Kalaallisut);Danish" tvg-logo="https://i.imgur.com/uQ50JUX.jpg" group-title="",KNR1 (1080p)
+https://iptv.dc-team.com/query?url=https://www.youtube.com/channel/UCIjOAtv53RHerlfl98fZmcQ/live
 #EXTINF:-1 tvg-id="KNR2.gl" tvg-country="GL" tvg-language="Greenlandic (Kalaallisut);Danish" tvg-logo="https://i.imgur.com/uQ50JUX.jpg" group-title="",KNR2 (720p) [Not 24/7]
-https://query-streamlink.herokuapp.com/iptv-query?streaming-ip=https://www.youtube.com/channel/UC6D225YIMCwVBiBktEeV-vQ/live
+https://iptv.dc-team.com/query?url=https://www.youtube.com/channel/UC6D225YIMCwVBiBktEeV-vQ/live
```
jkemp814, line 5: Could not get the link, Streamlink couldn't read https://www.youtube.com/channel/UC6D225YIMCwVBiBktEeV-vQ/live

Youtube: Video unavailable. The uploader has not made this video available in your country.

Must be [Geo-blocked] when viewed directly from YouTube.

dct-infra: Indeed geoblocked

Dum4G: @dct-infra I heard you are trying to implement proxy on your backend or a server switcher at least. Would it be able to use Greenland server or proxy? Thing is we have to trick YT that we're Greenland viewers, otherwise streamlink won't be able to grab the stream

dct-infra: DK proxies might work in general for geo restricted GL urls but are rare.
My implementation needs a reachable endpoint within the target country network to fully route to or partially to specific destination IPs/ranges via that endpoint. The magic happens at layer 3 which makes my implementation helpless in this specific case, at least for now.

LaneSh4d0w: Your implementation might work for something else : IP-locked links.

Dum4G: I think we don't need KNR2 link there but we can move on with KNR1 which isn't geo-blocked

Sphinxroot:
it was working at time when i saw it not it geoblocked by the user ..

PR #4878, opened 2021-10-03
moved a link

dct-infra:


jkemp814 (dismissed):
Good find. I think who added the channel thought it was Fashion TV (FTV)

But instead it's FTV Global News out of Republic of China (Taiwan)

BellezaEmporium (approved):

jkemp814 (approved):

PR #4959, opened 2021-10-09
EG : Add Quran TV

BellezaEmporium:
Quite confused by the origin of *this* specific channel. Got it called "Quran TV", but it isn't the Saudi version. I think it is the Iranian version, but it is streaming in Arabic, not in Persian. To confirm.

EDIT : As I've thought, the confusion took me and it also isn't the Iranian version. For those who know, please move this channel to it's correct place.

(Proof picture)

vlcsnap-2021-10-09-12h29m20s466


dct-infra:
Reading/reciting the Quran is always done in arabic and there is only one version no matter the country's official or spoken language.
This one can be from ir tr my id or any other non arabic speaking country, will be difficult to know.

jkemp814:
IP address from Spain

LaneSh4d0w:
> 
> 
> @LaneSh4d0w Reading/reciting the Quran is always done in arabic and there is only one version no matter the country's official or spoken language. After some searches the channel streams live on youtube, the owner is a renowned egyptian scholar, so channel is probably EG.

Nice. Moving it to EG.

LaneSh4d0w:
> 
> 
> IP address from Spain

Also right : https://search.censys.io/hosts/213.162.202.5

Dum4G (approved):

jkemp814 (approved):

PR #5056, opened 2021-10-16
Update us_pbs.m3u

what-a:
PBS moved their streams, so they were suddenly all broken. Here are the new locations.


MapGuy11 (changes requested):
Change all index_1.m3u8 to index.m3u8

MapGuy11 (approved):

freearhey (approved):

PR #5152, opened 2021-10-23
Update fr.m3u

mh292929:
remove 2 channels because channel names change


dct-infra (changes requested):
Thank you for your contribution.
Why would we remove WORKING channels? An update would be more relevant.
Feel free to update the channels according to their current state/name/logo/type but don't just remove.

freearhey (approved):

jkemp814 (approved):

dct-infra (approved):

PR #5169, opened 2021-10-24
Add More Hallmark Channels

MapGuy11:
1. Remove (1080p) badge (They actually max out at 720p)
2. Add Hallmark Drama
3. Add Hallmark Movies & Mysteries
4. Remove Geo-Blocked tags because they are not geo-blocked.
5. Remove User-Agent tags because I just didn't have the right key we don't actually need them.


freearhey (commented):
Doesn't work for me
image

freearhey (approved):

BellezaEmporium (approved):

PR #5236, opened 2021-10-30
Add new streaming link for Kapamilya Channel [Offline]

muxdigital:
Kapamilya Channel (new m3u8 link)


BellezaEmporium (changes requested):
We do not accept Xtream-Codes server links.

MapGuy11:
Doesn't Work anyway. We DONT accept any IPTV portal links!!!

MapGuy11 (changes requested):
Doesn't Work anyway. We DONT accept any IPTV portal links!!!

PR #5282, opened 2021-11-03
Update format.js

freearhey:
After this update, links that return a [403 HTTP](https://en.wikipedia.org/wiki/HTTP_403) code will be marked as `[Geo-blocked]`. At the moment such links are marked as `[Offline]` and excluded from public playlists.

Test results: https://github.com/iptv-org/iptv/commit/c2182962eb66fd558dbb13e89aec3e2c7d937771


LaneSh4d0w:
Quick notice, some might return 403 because they're tokened and might not work because of this.

freearhey:
If you know how to distinguish one from the other, let me know and I will take it into account in the code.

LaneSh4d0w:
> 
> 
> If you know how to distinguish one from the other, let me know and I will take it into account in the code.

As one example : if your script knows how to read the DOM, if you see a message around those lines "Streamer protection has denied the access to this content", it is the Flussonic stream manager asking a token, and so will return "Forbidden".

freearhey:
No, the script can't check that.

LaneSh4d0w:
> 
> 
> No, the script can't check that.

If you want, I can give it the ability to make verifications inside the DOM. Using node-fetch (a little program) and some regular expressions to look after certain words or patterns, we may put it onto the "403" switch condition.

freearhey:
Have you tested this code?

LaneSh4d0w:
> 
> 
> Have you tested this code?

I'm looking at it. Code should work if placed on "Offline", for verification.

BellezaEmporium (approved):

LaneSh4d0w:
Until I work on my solution to make it OK for production, i'll approve this.

dct-infra (approved):

PR #5290, opened 2021-11-03
GE : Scrape channels from Tulix

BellezaEmporium:


Dum4G (dismissed):

LaneSh4d0w:
Channels come from there : https://giniko.tv/package-channels.php?id=20

freearhey (changes requested):
None of the links work for me.
image

LaneSh4d0w:
Playlist retrieval works, AES Key retrieval blocks on certain programs, but works using browser extension...

Weird. Gonna draft this the time i'm searching for the solution.

LaneSh4d0w:
OK, after some fiddling I've found the issue. The AES server requires a session cookie to get the link, something that VLC and/or Streamlink do not gather over the requests, causing the AES server to constantly reject the requests.

I can inject cookies over headers before request redirection, thanks to my iptv-all server. I'll see what I can do.

PR #5321, opened 2021-11-07
Update ec.m3u

sguinetti:
This PR adds channels from http://eliteradioytelevision.com/index.php/tv-en-vivo/ and others


LaneSh4d0w:
Funny how they're not consistent in the names. I'll check the portalexpress.es server using my script a bit later, got other stuff running in the background;

BellezaEmporium (approved):

freearhey (approved):

sguinetti:
> Funny how they're not consistent in the names. I'll check the portalexpress.es server using my script a bit later, got other stuff running in the background;

That's true. I can't think of anything about how they updated the links to this type of url.

PR #5353, opened 2021-11-12
PE: Update CNC Digital and Chalaca TV

sguinetti:
This PR updates CNC Digital and Chalaca TV


taodicakhia (approved):

BellezaEmporium (approved):
For mediastreamm, it seems like only the port changes, like the portalexpress site. Can you confirm this ?

sguinetti:
> mediastreamm

Maybe no. In this channel https://mediastreamm.com:3162/stream/play.m3u8 no works but multi_web yes. Other links from mediastream like QT works only with */stream/* but no https://mediastreamm.com:3753/multi_web/play.m3u8.

freearhey (approved):

PR #5361, opened 2021-11-13
EC: Add EducaTV

sguinetti:
This PR adds EducaTV from https://www.educa.ec/


BellezaEmporium (dismissed):

dct-infra (commented):
Takes ages to load and then it randomly drops, probably the round trip time, needs another review than mine.

sguinetti:
> Takes ages to load and then it randomly drops, probably the round trip time, needs another review than mine.

They said this channel would be geo-blocked. I don't know if the slow response out of the country of the stream is related to this.

LaneSh4d0w:
Long connection time, not geo-blocked from where I am, but works. Poor bandwidth over server, but works.

BellezaEmporium (approved):

freearhey (approved):

PR #5386, opened 2021-11-15
CR: Add Enlace TV

sguinetti:
This PR adds a channel from https://www.enlace.org/. But the stream is taken from https://www.enlacebpb.cl/, because the original stream is broken.


dct-infra:
according to lyngsat this channel is .cr, can you confirm/check?

sguinetti:
Right. In Wikipedia, the channel streams two version TBN-Enlace and Enlace https://en.wikipedia.org/wiki/Enlace. The second is taken for this PR. I will edit.

BellezaEmporium (approved):

dct-infra (approved):

PR #5400, opened 2021-11-17
CR - Testing : Add alternative Zona Films + Zona Terror

BellezaEmporium:


dct-infra (commented):
does not load for me, might be the flv stuff.

LaneSh4d0w:
> 
> 
> does not load for me, might be the flv stuff.

Most certainly, it is weird how it works on VaughnLive but VLC only sees a pack of bytes for no reason.

PR #5437, opened 2021-11-21
JP : The "Unexpected, but it is here" commit.

BellezaEmporium:
Redlab left, long live Redlab !

See #5433

jp


Dum4G:
I think it's better to find an old revision of jp playlist and just change links with batch replacement to preserve channel tags. I'll do that at some point.

LaneSh4d0w:
> I think it's better to find an old revision of jp playlist and just change links with batch replacement to preserve channel tags. I'll do that at some point.

You're right, let me check that out.

LaneSh4d0w:
@Dum4G Done, please check to make sure everything's in order.

Dum4G (dismissed):

Dum4G (approved):

freearhey (approved):

PR #5584, opened 2021-12-05
UY: Update Latino Kids

sguinetti:
The PR uses a aditional url from m3u.cl.


Dum4G:
Is it something that belongs to them or specifically hosted for them?

LaneSh4d0w:
Well... I must say I don't understand why the additional query parameter would lock/unlock anything.

sguinetti:
The url stream is now inaccessible with strict origin, but addding parameter would make the stream works without problems.

LaneSh4d0w:
I personally don't have issues with the original link. What's your player ?

sguinetti:
> I personally don't have issues with the original link. What's your player ?

VLC, yesterday.

sguinetti:
I close for now. However, for this possible problem, this solution may be invoked.

PR #5711, opened 2021-12-14
Create xxx.m3u

MKOUYT566:
https://iptv-org.github.io/iptv/categories/xxx.m3u


sguinetti:
There is no need to create a file, the link is generated automatically. Thank you.

abofresa:
Thank you.

PR #5761, opened 2021-12-16
SA : Change link for Ahlu Al-Quran

BellezaEmporium:
While I'm against Xtream-Codes usages, it seems like they use one for restreaming their channel.

https://www.ahlu.tv/live-tv/


Dum4G:
KEKW

Dum4G (approved):

dct-infra (approved):

PR #5880, opened 2021-12-26
KR : ???

BellezaEmporium:
Seems that this channel stopped geo-locking ?

Tested with Serbian IP :
ko

Server seems to be a lot more stable and having a better bandwidth in the same time.


dct-infra (approved):
tested, works from Russia, not from EU

PR #5881, opened 2021-12-26
JP : ???

BellezaEmporium:
Same for this one

an

Tested with Serbian IP.


Dum4G:
We had many complaints before that the source is not available in Indonesia itself

LaneSh4d0w:
Ah, they reverse-carded then. Before it wasn't available to us and now it's
not available to Indonesia. Weird.

dct-infra (commented):
Same comment as for KR

PR #5920, opened 2021-12-31
Update regions.json

freearhey:
- added missing countries to the MENA region
- added [Northern America](https://en.wikipedia.org/wiki/Northern_America) region (Resolves #5624)
- added [Latin America and the Caribbean](https://en.wikipedia.org/wiki/Latin_America_and_the_Caribbean) region


Sphinxroot (approved):

scripts/data/regions.json
```diff
@@ -561,10 +612,12 @@
     "country_codes": [
       "AE",
       "BH",
+      "CY",
       "DJ",
       "DZ",
       "EG",
       "EH",
+      "IL",
```
LaneSh4d0w, line 620: Israel ? Not quite sure about this one.

scripts/data/regions.json
```diff
@@ -561,10 +612,12 @@
     "country_codes": [
       "AE",
       "BH",
+      "CY",
```
LaneSh4d0w, line 615: Cyprus ? Not quite sure about that one.

scripts/data/regions.json
Sphinxroot: https://www.iban.com/country-codes

LaneSh4d0w: Yes, that confirms what I was saying. IL is Israel, it is not a country around the LATAM region, but on the Middle East one.

freearhey: Please check again. These changes have nothing to do with the LATAM region.

BellezaEmporium (approved):

PR #6045, opened 2022-01-09
Corrected spelling of language

saurabhperiwal:
'Panjabi' to 'Punjabi'


Dum4G:
Won't change anything. We need to find every channel where the language is set and change it there

LaneSh4d0w:
Both prononciations seems to be correct according to Wikipedia : https://en.wikipedia.org/wiki/Punjabi_language

Issue #6224, opened 2022-01-29
Replace: All links at "50.7.74.29" 


lezhangisu:
### Channel Name

Japan TV channels

### What happened to the stream?

It is buffering to unplayable point

### Playlist

https://github.com/iptv-org/iptv/channels/jp.m3u

### Broken Link

All channels directed to ip 50.7.74.29

### Possible Replacement

NA

### Notes

The ip 50.7.74.29 is not ping-able now. It was good 5 hours ago. Is the server down? Anyone can fix it? Thanks

Central302:
[Memo] All Port Not Work[Tine-Out]

pencail:
Yes, the animax that I love to watch can't be watched either。ping 50.7.74.29，prompt timeout

iant1000:
All Japanese channels with 50.7.74.29 are still not working. It would be amazing if you could fix it. Thanks in advance

pShota:
drop my 2 cents confirm all 50.7.74.29 channels are dead

lezhangisu:
I am able to ping the server 50.7.74.29 today. But streaming services still not available.

Nobodynumber1:
Before had hacker to down services?

Today services had run again.

[Feb 02, 2022 11:43 PM UTC ]
Location Updated
[Feb 02, 2022 11:43 PM UTC ]
Routing Updated
[Feb 02, 2022 11:43 PM UTC ]
Orange
Service Added
 22/TCP/SSH

UnderPM:
Stopped working again about 9h ago

PR #6227, opened 2022-01-29
Update JP.m3u (Goodbye 50.7.74.29)

Central302:
This Patch Will Apply #6224 .
Server No Longer Response about ping. So,I Thought No Hope to FIX Server
Thank You 50.7.74.29 .


Dum4G:
It has restored once, let's wait and see

LaneSh4d0w:
Officially offline.

We haven't found any publicly accessible services on this host or the host is on our blocklist.

PR #6236, opened 2022-01-29
Remove Not 24/7 these are fixed with multibitrate

MapGuy11:
found a new way


Dum4G (approved):

what-a (approved):
Relative URLs to the rescue.

PR #6321, opened 2022-02-02
Fix NY LocalBTV

MapGuy11:


LaneSh4d0w:
Draft ?

MapGuy11:
> Draft ?
> 
> * AZ AM 731 CA using same NY method works too. (nvm, works on all channels / all servers, same for TX 873 FN).

No, I just need to fix one thing.

what-a (approved):

BellezaEmporium (approved):

Issue #6429, opened 2022-02-15
Fix: All the link related with 50.7.74.29


luongz:
### Channel Name

NHK G and more (related to 50.7.74.29)

### What happened to the stream?

It is buffering to unplayable point

### Playlist

https://iptv-org.github.io/iptv/countries/jp.m3u

### Broken Link

http://50.7.74.29:8880/hls/j00033/index.m3u8

### Possible Replacement

Change the IP and Port to: 51.79.193.108:8080

Nobodynumber1:
**Too Strong 
Fixed Got new IP**
Thanks

Central302:
Fixed. I'll Close The Issue.

PR #6430, opened 2022-02-15
Update jp.m3u

Central302:
This Patch Will Apply
●Fix of #6429 (New IP Adress Thanks By @luongz)


butatuli (approved):
Really changing the ip is the solution to these, thank you

GarckaMan (approved):
All offline jp channels are back with the new IP. Thanks!

freearhey (approved):

Dum4G (approved):

PR #6477, opened 2022-02-21
Reflect changes from 02.2022 onto README

BellezaEmporium:


freearhey:
There is no need for this, the readme will be updated later.

PR #6702, opened 2022-03-14
Add country subdivisions codes to the country lists

gabrielbergeron:
I found out that some of the channels I usually watch disappeared a few weeks/months a go from the CA list. I found out that it's because the channels broadcast_area where marked to subdivisions instead of the country code.
I figured we could also load the channels from the subdivisions from the countries when creating the countries playlists.


LaneSh4d0w:
Assigning Arhey on this PR, since it's quite a changing move, adding something more to countries.

freearhey:
@gabrielbergeron Please add tests to make it clear what results you expect to see.

freearhey (dismissed):

freearhey:
@gabrielbergeron Please check the playlists again, the channels should already appear.

gabrielbergeron:
@freearhey it works great, thank you! Also thank you for adding the tests for me, I didn't have the time since I was working.

PR #6721, opened 2022-03-15
set [Geo-blocked] for a geo-blocked channel

Owyn:
streamtest.in says it's offline but for me it always works as I've checked for multiple past days, and channels with similar URL already have [Geo-blocked] status


Owyn:
also, it's not properly saying "blocked" for 451 error it seems
изображение (tried opening it via proxy)

freearhey (approved):

LaneSh4d0w:
451 indeed.

BellezaEmporium (approved):

Issue #6843, opened 2022-03-25
Fix: add error 451 alongside error 403 for "blocked" status to scripts


Owyn:
### Describe your issue

> 451 means that content is not available due to legal reasons. Another way of saying "geo-locked".

https://github.com/iptv-org/iptv/pull/3883#issuecomment-899639090
https://github.com/iptv-org/iptv/pull/6839#pullrequestreview-920888536

![image](https://user-images.githubusercontent.com/1309656/160094989-7954147f-e447-465b-81ca-20e90cc86b20.png)

PR #6848, opened 2022-03-25
fix blocked status for geo-blocked content 451 #6843

Owyn:


its0x08 (approved):

freearhey:
We cannot mark all 4xx errors as blocking, because most of them have a different meaning: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#client_error_responses

PR #6887, opened 2022-03-29
Update ve.m3u

sguinetti:
This add new streams from Venezuela


freearhey (approved):

dct-infra (approved):

BellezaEmporium (changes requested):
Didn't we say we do not accept accents in tvg-id ?

freearhey:
@LaneSh4d0w you are right, the correct id is `MundoMas.co`. I'm going to correct that now.

LaneSh4d0w:
> @LaneSh4d0w you are right, the correct id is `MundoMas.co`. I'm going to correct that now.

Moreover, this link is already available in mx.m3u.

freearhey:
@LaneSh4d0w I removed that link.

PR #7148, opened 2022-04-25
FR : Fix and add some channels

BellezaEmporium:
I won't tell anything about the source, apart that it's an Astra server.


dct-infra (approved):
nice!

its0x08 (approved):

what-a (approved):

PR #7150, opened 2022-04-25
Update id.m3u

F640:
- Added Magna Channel and Metro Globe Network from [medcom.id](http://medcom.id) 
- Added better and (possibly not geo-blocked) Kompas TV URL, 480p and 360p variant also available by changing `kompastv_lhd.m3u8` to `kompastv_lsd.m3u8` or `kompastv_lld.m3u8`. Kompas TV live also available from [YouTube](https://m.youtube.com/c/kompastv/live). Btw links are from [here](https://www.kompas.tv/live2).
- Added YouTube URL for NET., Metro TV, and TVOne. TVOne needs to change its link after livestream ends as the link is unlisted for unknown reason and new link can be taken from [here](https://www.tvonenews.com/live)
- Replace The Indonesia Channel link to working ones
- Removed Radio 51 TV (Italian Channel)
- Changed some channel IDs


F640:
> It seems there's also https://live-kg.jixie.media/live/kompastv.m3u8

That's unexpected, but it has no difference, it only uses best quality.

LaneSh4d0w:
> > It seems there's also https://live-kg.jixie.media/live/kompastv.m3u8
> 
> That's unexpected, but it has no difference, it only uses best quality.

The only difference is that the YouTube stream is different from this one it seems.

F640:
But it's not a problem anyways, they both already good for now.

BellezaEmporium (approved):

its0x08 (approved):

what-a (approved):

PR #7152, opened 2022-04-25
US_Pluto : EPG

BellezaEmporium:
kudos and thanks to @matthuisman for this.

(If we keep this EPG, tvg-ids will be changed accordingly. This is a proposition from an already existent source that is available freely thanks to the mentioned user above)


its0x08 (approved):

freearhey (commented):
I've already added this guide to the [iptv-org/epg](https://github.com/iptv-org/epg)

```
https://iptv-org.github.io/epg/guides/us-pluto/i.mjh.nz.epg.xml
```

so updating the playlist is no longer necessary.

Also, the x-tvg-url value from internal playlists is not used when generating public playlists.

PR #7245, opened 2022-05-05
NSFW channels appearing in index.m3u instead of index.nsfw.m3u

mrwhizzy:
Multiple NSFW streams had been added in `streams/cy.m3u` which are not relevant to any Cypriot channels.


its0x08 (approved):

mrwhizzy:
I get that these channels might be Cyprus-originated, however, they seem appear in `https://iptv-org.github.io/iptv/index.m3u` as well.

From what I understand, this list should not contain NSFW channels and that these channels should appear in the list that includes adult channels `https://iptv-org.github.io/iptv/index.nsfw.m3u`.

LaneSh4d0w:
> I get that these channels might be Cyprus-originated, however, they seem appear in `https://iptv-org.github.io/iptv/index.m3u` as well.
> 
> From what I understand, this list should not contain NSFW channels and that these channels should appear in the list that includes adult channels `https://iptv-org.github.io/iptv/index.nsfw.m3u`.

Now, that's an issue ! I'm going to verify onto iptv-org/database, see if we missed those ones.

LaneSh4d0w:
I'm making the necessary changes to database, i'll close this PR, and should be fixed on next playlist update.

PR #7305, opened 2022-05-17
Add Iran TV Channels

dev-onetv:
This PR adds back the Iran TV channels that  previously added to `ir_telewebion.m3u` in #7280. That file somehow got deleted with no information!


LaneSh4d0w:
Had made a PR about deleting these links since they used tokenized versions. Never seen those links before.

dev-onetv:
The links point to the Telewebion official website. These links do not have tokens, but their server includes token for the tracks of the playlist. I will close the PR anyway. Thanks

PR #7387, opened 2022-05-29
Update to the Official Stream Link

VoXaN24:


LaneSh4d0w:
It would work if there wasn't any token. sadly it always changes. The only way it could work is basically creating a Streamlink plugin for TegoTV.

its0x08 (approved):

taodicakhia:
> It would work if there wasn't any token. sadly it always changes. The only way it could work is basically creating a Streamlink plugin for TegoTV.

Streamlink doesn't support TegoTV stream, and token of stream contains IP so you can't play stream in other IP

LaneSh4d0w:
> > It would work if there wasn't any token. sadly it always changes. The only way it could work is basically creating a Streamlink plugin for TegoTV.
> 
> Streamlink doesn't support TegoTV stream, and token of stream contains IP so you can't play stream in other IP

Good point. It would only work locally then.

LaneSh4d0w:
Cannot be added to the playlists, due to @taodicakhia's answer mentioning the tokens being IP-locked.

PR #7563, opened 2022-06-21
CO : Reinstate the hvtrafico links

BellezaEmporium:
After a severe downtime, it seems they came back.

We'll need to have a deep look over Cinema+ around the 23:00 / 5:00 period, as it might be like Rouge TV : stream family friendly content over daytime, and 18+ content over night time.

(it seems they're highly unstable. they're using DDNS, most certainly is a self-hosted stream (computer or server)


its0x08 (dismissed):

what-a:
None work for me. Geo-blocked or perhaps down?

LaneSh4d0w:
It seems their streams are highly unstable. Used to have them working last time.

LaneSh4d0w:
Confirmed, Cinema+ having 18+ content, are affiliated with Colombian X studio https://www.switchpremium.com/

LaneSh4d0w:
The weirdest thing is : the server isn't even **down**, it answers to pings but fails to give proper response.

LaneSh4d0w:
It is now working... for now.

LaneSh4d0w:
Closing since it seems they didn't came back over. Not removing the branch because if they do appear again (and are stable), we could pull them over.

PR #7590, opened 2022-06-25
Update vn.m3u

kirigizakazutovn:
Re-update some Vietnamese channel
It's work,I tried that


BellezaEmporium (changes requested):
Ouch. Lots of things are wrong in there.

- No advertisements are tolerated inside the files.
- Adaptive playlists instead of chunklists
- Do **not** remove working links, add alternatives or change them if they aren't working.
- Sounds like it's a copy and paste of another repository.

LaneSh4d0w:
Closed due to lack of answer by OP.

PR #7839, opened 2022-07-23
KR : Add specific quality

BellezaEmporium:
Prevent having only one link available when there's 3 of them for each.


LaneSh4d0w:
Another server : https://apichvod33.seezntv.com/

freearhey (commented):
Titles are not used when filtering links.

LaneSh4d0w:
> Titles are not used when filtering links.

Then it's the tvg-id part... that I can't change due to the EPG.

freearhey:
I guess so.

PR #7847, opened 2022-07-24
Update mw.m3u

BellezaEmporium:


its0x08 (dismissed):

freearhey (approved):

benevenstanciano:
This should be Mauritius (.mv) rather than Malawi (.mw) right?

freearhey (changes requested):
https://github.com/iptv-org/iptv/pull/7847#issuecomment-1193737336

LaneSh4d0w:
> This should be Maldives (.mv) rather than Malawi (.mw) right?

Indeed, the little difference fooled me.

freearhey (approved):

what-a (approved):

PR #8038, opened 2022-08-14
Fix CanalAntigua.gt and TN23.gt channels

aalonzolu:


sguinetti:
I keep old Canal Antigua stream becuase it is online yet. You do not need to replace a online stream for add other.

PR #8321, opened 2022-08-27
Update fr_samsung.m3u

Sphinxroot:


UltraHDR (approved):

UltraHDR (approved):

BellezaEmporium (commented):
Shouldn't the rakuten ones be on the fr_rakuten.m3u file instead of here ?

LaneSh4d0w:
Yeah that's where it's tricky... but most of the times if you get them from the "Rakuten TV" app, then it's most certainly this.

BellezaEmporium (approved):

PR #8323, opened 2022-08-27
Update ie_samsung.m3u

Sphinxroot:


BellezaEmporium (commented):
AVF -> AFV (America's Funniest Videos)

same than the other ones, I believe the Rakuten ones should belong in a at_rakuten.m3u file.

PR #8324, opened 2022-08-27
Update be_samsung.m3u

Sphinxroot:


BellezaEmporium (commented):
Didn't know there was a Belgian Rakuten TV. I believe we should put them onto a be_rakuten.m3u file, following the French one.

what-a (approved):

UltraHDR (approved):

PR #8375, opened 2022-08-30
Update in.m3u

apnatvuk:
please  add this channel


UltraHDR:
Channel is UK based http://apnatv.uk/ and will be added in https://github.com/iptv-org/iptv/pull/8376

PR #8432, opened 2022-09-01
WIP : Split Pluto.tv links into different playlists

BellezaEmporium:
Made over an undivided approval (https://github.com/iptv-org/iptv/discussions/8332), will follow the same path as the Rakuten links.

NOTE : Some channels won't/can't be moved since I have literally no data over them (look at the number of "Pluto.tv MTV" or other sorts of channels as such.
Some channels that have a "in Spanish" channel, unless being said from Spain, aren't in the Spanish playlist, since it can also be from Mexico (see that as LATAM).


LaneSh4d0w:
Now R2R. I've moved the most obvious channels. Now let's play hide and seek with the already known URLs to see which US channels correspond to which countries... What a pleasure.

freearhey (approved):

Alstruit (approved):

PR #8836, opened 2022-09-18
TL: creation & add RTTL East Timor

BellezaEmporium:
Not 24/7, seems to stream every hour.

Has Tetum, Portuguese and Indonesian news panels.


UltraHDR:
Hasn't streamed in 8 hours but I'll approve anyway

UltraHDR (approved):

freearhey (commented):
The original link works:
https://www.youtube.com/channel/UCtwMP_IzEm3r9PDuaYRUKLA/live
image

Streamlink does not.

LaneSh4d0w:
> The original link works: https://www.youtube.com/channel/UCtwMP_IzEm3r9PDuaYRUKLA/live image
> 
> Streamlink does not

The channel doesn't stream all the time, it seems they do get online once it's news or anything related to it.

freearhey:
The problem is that the link does not work even when the source is online.

LaneSh4d0w:
> The problem is that the link does not work even when the source is online.

That's odd, I'll try when it comes back online.

UltraHDR:
> The problem is that the link does not work even when the source is online.

It worked for me in VLC

UltraHDR (approved):

freearhey (approved):
I just checked again:

- without VPN via VLC, it worked for me too
- with US VPN still gives me an error

But maybe it's just a problem on my end.

PR #8885, opened 2022-09-20
Update sd.m3u

UltraHDR:


BellezaEmporium (approved):

freearhey (commented):
The `status` attribute is updated automatically during daily checks, there is no need to update it manually.

LaneSh4d0w:
> The `status` attribute is updated automatically during daily checks, there is no need to update it manually.

I was thinking maybe we could port our channel verification trigger up to the PRs too. It could render a message telling if the channel is online / offline. Thus we can easily get a geo-block or if it's an issue server or script-related.

freearhey:
@LaneSh4d0w This is a good idea! I can't do it myself yet, but you can post it in the Discussions and maybe someone will implement it sooner.

UltraHDR:
> The `status` attribute is updated automatically during daily checks, there is no need to update it manually.

Thanks. I'm aware. I noticed it was always wrong.. Is it online for you now?

freearhey:
> Is it online for you now?

Yes.

PR #9038, opened 2022-09-25
Update uk.m3u

hypetvuk:
hype tv live from the streets of southall london

PR #9039, opened 2022-09-25
Update uk_samsung.m3u

hypetvuk:
hype tv


UltraHDR:
Already added in https://github.com/iptv-org/iptv/pull/9038

PR #9040, opened 2022-09-25
Update uk_channelbox.m3u

hypetvuk:
hype tv live from uk


UltraHDR:
Already added in https://github.com/iptv-org/iptv/pull/9038

PR #9328, opened 2022-10-03
Adding Animax Update ph.m3u

Rocsondaboss:


LaneSh4d0w:
already available here : https://github.com/iptv-org/iptv/blob/d034198ffe192c639a5ed174e45f98aef6a2d36c/streams/vn_fptplay.m3u#L2-L7

PR #9391, opened 2022-10-04
Update hu.m3u

kristoft13:
Added M4 Sport (720p)


LaneSh4d0w:
Due to the nature of the channel, I don't think it'll be accepted. @freearhey ?

benevenstanciano:
The channel has two versions - one that's available across Europe and blacks out international sport, and one internal version. The first would probably be fine. In any case, looks like the stream isn't working.

LaneSh4d0w:
> The channel has two versions - one that's available across Europe and blacks out international sport, and one internal version. The first would probably be fine. In any case, looks like the stream isn't working.

Well the last time i've tested it streamed live matches of the UEFA Champions League.

LaneSh4d0w:
Not working anymore, and due to no answer from OP since then + no change, closing this PR.

PR #9572, opened 2022-10-15
Update EnerGeek URLs

EnerGeekTeam:
Updated URLs for EnerGeek channels.


sguinetti:
Thank you for your collaboration. In general, when there are several channels broadcasting differently, we set different ids for each stream. For now we use the ids from Database but if you need to edit some information you can let us know in the other repository. See https://github.com/iptv-org/iptv#database.

EnerGeekTeam:
I think all 4 of our channels should use the same ID as they are all from our brand.

sguinetti:
No. The ids are almost similar to those of CCTV, for example. There is an id called CCTV1.cn for one channel, CCTV2.cn for another, and so on. I admit that I created those ids to distinguish Energeek 2 and Energeek 3 programming because I have not confirmed their official name. So I ask, could you confirm that the channels 2 and 3 are permanently named "Energeek Fanpop" and "Energeek Wowly" for editing in the database, please?

EnerGeekTeam:
At least for EnerGeek 2, it is permanently renamed to EnerGeek Fanpop. EnerGeek 3 is a temporal alliance we created with another brand.

sguinetti:
Thank you. For now, I'll let the ids of the streams you shared remain as "Energeek1.cl", "Energeek2.cl" and "Energeek3.cl". Here you can review the information for each of the ids on this website: https://iptv-org.github.io/?q=Energeek 
I will merge these changes, but in case of any significant change with the ids it will be done in another PR.

sguinetti (approved):

Alstruit (approved):

PR #9632, opened 2022-10-18
Update cl.m3u

sguinetti:
Rename id of Energeek2 (now Fanpop), see https://github.com/iptv-org/database/issues/765.


BellezaEmporium (approved):

EnerGeekTeam:
Was about to create a PR because EnerGeek Fanpop appears on the website but no streams are attached to it.

sguinetti:
Right. When an id is changed in the Database repo, it must also be changed immediately in the playlist. It is necessary for the streams to appear with the new id as possible. As you may understand, this PR requires an admin approval but it won't take long. We appreciate your patience and understanding. Additionally, the whole process is also explained in the FAQ page.

sguinetti:
Feel free to update other channel details in the Database as needed. This will not affect the streams of the playlist generated on the web, except when the id is changed. As this is the first time an owner has done this process, we are happy to receive feedback.

Alstruit (approved):

EnerGeekTeam:
Well i'm not the founder of the project itself but i am the one that works IT here.

PR #10243, opened 2022-11-11
Update tr.m3u8

kaypoush:


BellezaEmporium (changes requested):
Per data from https://yoltv.eu, Turkish diaspora channel from Köln, Germany. Please move it to the de.m3u playlist.

kaypoush:
AFAIK Yol TV is an opposition channel and was banned by the Turkish government so they are operating from Germany. Their programs are in Turkish and focused on Turkey rather than Germany. IMO it would be weird for the Germans to see this channel in their list. Do you still think I should move it to de.m3u?

LaneSh4d0w:
If it has its roots in Turkey then I apologize, i'll approve it.

BellezaEmporium (approved):

UltraHDR (approved):

PR #10311, opened 2022-11-13
Update br.m3u

sguinetti:
This PR readd Peru Channel


BellezaEmporium (approved):
Not entirely sure from where it's from but right now it airs some multi-lingual program linked to religion ?

sguinetti:
Right, Is a broadcast syndication. But it does not mean that the name of the channel will change.

LaneSh4d0w:
Was more thinking about the country of origin, but yes if it's a return and nothing has changed we'll change nothing in these parts.

Alstruit (approved):

PR #10809, opened 2022-12-14
Added some local broadcasters

KDra:
Added the public broadcaster and a local channel


BellezaEmporium (commented):
tvg-logo, group-title aren't used anymore on these spots. There's no need to have them.
tvg-id is not one of the links that is available in the database.

LaneSh4d0w:
Closing due to lack of answers by OP.

PR #10983, opened 2022-12-25
updated source for ph channels

LouiseDenmar:


LaneSh4d0w:
Manually changing the status is not useful, because it'll be changed again when the bot kicks in.

PR #11050, opened 2022-12-29
Change Samsung ES

joaquinito2070:


LaneSh4d0w:
while it is not wrong *per-se* this is a whole copy and paste of @matthuisman 's hard work and does not follow our notation guidelines.

PR #11095, opened 2023-01-02
Update es.m3u

trincowski:
- Added new channels
- "TV3Cat.es" has 3 different links, since depending on the VPN server, some might work while others won't


trincowski:
Closing PR to resolve some issues.

PR #11119, opened 2023-01-05
Update pe.m3u

sguinetti:
This PR is for new channels.


LaneSh4d0w:
The PBO stream may be tokened somehow someway, the "ey" makes me think that's an JWT token, and could change anytime.

Here's the base data for the stream : https://player-backend.restream.io/public/videos/60098b6e980048d09346e9589af6154f

sguinetti:
> The PBO stream may be tokened somehow someway, the "ey" makes me think that's an JWT token, and could change anytime.
> 
> Here's the base data for the stream : https://player-backend.restream.io/public/videos/60098b6e980048d09346e9589af6154f
> 
> image

I did not expect. I found from blob:https://player.restream.io/bbb4ce0c-5f14-42d3-8d30-8af35edacda9 On the website source code I found the apparent permanent url.

BellezaEmporium (approved):

UltraHDR (approved):

PR #11227, opened 2023-01-14
Update de.m3u

francisuk1989:
English https://club.54house.fm/en/
German https://club.54house.fm/de/
Song title and artist will display if the receiver supports it.


UltraHDR:
Thnaks for your contribution.
Sorry. We only accept links with Video.

For Radio see: https://www.radio-browser.info

PR #11436, opened 2023-01-25
Update us.m3u

trincowski:
Add New Source for "Journy.us"
- The one in "us_xumo.m3u" seems to be linked to a Bear Grills show. **Since they are different Streams, maybe a new tag-id should be used for the other channel?**


LaneSh4d0w:
If the other one's a Bear Grylls stream, then we should check if it's the correct stream or not.

trincowski:
> If the other one's a Bear Grylls stream, then we should check if it's the correct stream or not.

Well, it must be a VOD program, since I grabbed this stream from the site itself: https://journy.tv/watch/?live=journy

This one is the Live Stream.

LaneSh4d0w:
> > If the other one's a Bear Grylls stream, then we should check if it's the correct stream or not.
> 
> Well, it must be a VOD program, since I grabbed this stream from the site itself: https://journy.tv/watch/?live=journy
> 
> This one is the Live Stream.

indeed. just checked and it corresponds to your stream.

BellezaEmporium (approved):

PopeyeTheSai10r (approved):

PR #11469, opened 2023-01-29
Update gt.m3u

trincowski:
- Add new and faster Streams for a few existing Channels
- Add 15 new Channels
- Correct some names to make them consistent with the others
- Add Resolution and "[Not 24/7]" to some Channels

**Notes:**
- Streams from "http://204.12.202.90" server are all official and were grabbed from here: https://www.playgt.net/instrucciones
- None of these channels are Geo-Blocked, although they throw a 403 Error when they are not streaming


LaneSh4d0w:
> * Add new and faster Streams for a few existing Channels
> 
>     * Add 15 new Channels
> 
>     * Correct some names to make them consistent with the others
> 
>     * Add Resolution and "[Not 24/7]" to some Channels
> 
> 
> **Notes:**
> 
>     * Streams from "http://204.12.202.90" server are all official and were grabbed from here: https://www.playgt.net/instrucciones
> 
>     * None of these channels are Geo-Blocked, although they throw a 403 Error when they are not streaming

Aaaaah, Xtream-Codes when you hold us... lol. As long as it's official I won't refuse it.

BellezaEmporium (approved):

PopeyeTheSai10r (approved):

PR #11592, opened 2023-02-03
Fix timeout issue

freearhey:
On December 8, 2022, for some reason the [FFmpeg](https://ffmpeg.org/)  library that we use to check streams started to crash with error:

```sh
Process completed with exit code 139.
```

This error was incorrectly interpreted by the bot as  "FFMPEG Process Timeout":

[https://github.com/iptv-org/iptv/actions/runs/3648283497/jobs/6161528091](https://github.com/iptv-org/iptv/actions/runs/3648283497/jobs/6161528091)

As a result, all working links were marked as `timeout` instead of `online`.

As I found out this error occurs only with FFmpeg installed via the [FedericoCarboni/setup-ffmpeg](https://github.com/FedericoCarboni/setup-ffmpeg) action. If you install it directly, it works fine. So I updated the [auto-update](https://github.com/iptv-org/iptv/commit/7025f6cebd35f41e9b299d6c259a33a3584882a7) workflow and now the installation will go via [apt](https://en.wikipedia.org/wiki/APT_(software)).

Also now if a similar error occurs the bot will leave the status of the link unchanged.

Test run: [https://github.com/iptv-org/iptv/actions/runs/4080845738](https://github.com/iptv-org/iptv/actions/runs/4080845738)


BellezaEmporium (approved):

LaneSh4d0w:
Could it be linked with the channels disappearing for no reason while getting posted on the playlists ?

freearhey:
@LaneSh4d0w Actually, it should not have had any effect on the links themselves, since the bot only removes those that have the status of "error". So it is unlikely.

PopeyeTheSai10r (approved):

freearhey:
@LaneSh4d0w Just in case, you can always find a list of all the links removed by the bot and what exactly was the reason through the log:

https://github.com/iptv-org/iptv/actions/runs/4093304575/jobs/7059329301

image

PR #12505, opened 2023-04-23
Use HTTPS instead of HTTP m3u URLs for enhanced security and privacy

PeterDaveHello:
In 2023, HTTP is no longer secure and should be replaced by HTTPS whenever possible. This ensures not only safety but also privacy in all online activities.

Those http URLs being replaced with https protocal, were tested that their servers support https communication properly before the replace.


BellezaEmporium (approved):
LGTM

freearhey (approved):

PR #12644, opened 2023-04-30
Format internal playlists

freearhey:
Changes:

- links sorted alphabetically
- removed the `status` attribute that is no longer in use


BellezaEmporium (approved):

PopeyeTheSai10r (approved):

mpbejo:
Hi to all, in the past I used in a my app the status="online" to filter the working streams.
Only now I see that You don't use anymore.
So my question is, how can I filter the working streams now?
Thank You
Best regards
Marco

BellezaEmporium:
> Hi to all, in the past I used in a my app the status="online" to filter the working streams.
> Only now I see that You don't use anymore.
> So my question is, how can I filter the working streams now?
> Thank You
> Best regards
> Marco

The only way to make sure they work is to actually check the status code of the stream itself.

BellezaEmporium:
The ideal situation would be this, if you plan to take care of all streams.

mpbejo:
Well I don't know the reason to remove STATUS, but in my case it was much faster to understand if a stream was correct or not.
I explain, I developed a free app where a user can choose a country and from https://iptv-org.github.io/api/channels.json
I select all the id with the correct country and then from https://iptv-org.github.io/api/streams.json I select only the channels with the status online, but now without this parameter I think the selection of all the online channels becomes extremely slow.

PR #13225, opened 2023-06-14
Update kw.m3u

UltraHDR:


freearhey (commented):
Here's what I see:

image

Is it supposed to be like this?

UltraHDR:
Sorry I didn't realise the url was already in the file. I don't get your error BTW.

PR #13561, opened 2023-07-17
Update ve.m3u

ElvisGarces:
edit InterTV name


freearhey:
According to the record in the [iptv-org/database](https://github.com/iptv-org/database), the channel name is already correct: https://iptv-org.github.io/channel?id=InterTV.ve

PR #13687, opened 2023-07-29
Update ve.m3u

ElvisGarces:
Modify the name of Inter TV to InterTV, eliminating the space


freearhey:
According to the record in the [iptv-org/database](https://github.com/iptv-org/database), the channel name is already correct: https://iptv-org.github.io/channel?id=InterTV.ve

ElvisGarces:
The error is in channel name but no in channel ID, please correct this is
value is InterTV (1080) for search parameter.

PR #13696, opened 2023-07-31
Update ve.m3u

ElvisGarces:
Change the name of the channel from Inter TV (with a space) to InterTV (without a space), not in the ID which is well written without a space, but in the name that goes after the comma. This change is motivated by the fact that Inter TV already exists (with space) which belongs to the Inter Milan soccer team. Thank you.


freearhey:
> According to the record in the [iptv-org/database](https://github.com/iptv-org/database), the channel name is already correct: https://iptv-org.github.io/channel?id=InterTV.ve

https://github.com/iptv-org/iptv/pull/13687#issuecomment-1656827505

PR #13960, opened 2023-08-18
Update iq.m3u

Redo-San:
Added more Iraqi channels.

### DIFF
--- a/streams/iq.m3u
+++ b/streams/iq.m3u
@@ -153,7 +153,7 @@ https://livestream.almanasa.tv:8117/islamic/ch3/ch3_1080.m3u8
 https://stream.veo.buzz:6050/islamic/ch5/ch5_720.m3u8
 #EXTINF:-1 tvg-id="KurdChannel.iq",Kurd Channel (480p)
 https://kurd-channel.ikoflix.com/hls/stream_2.m3u8
-#EXTINF:-1 tvg-id="kmcTV.iq",KMC TV - Kurdish Music Channel TV (720p)
+#EXTINF:-1 tvg-id="kmcTV.iq",KMC TV Kurdish Music Channel TV (720p)
 https://kmchls.wns.live/hls/stream.m3u8
 #EXTINF:-1 tvg-id="Kurdistan24.iq",Kurdistan 24 (720p)
 https://d1x82nydcxndze.cloudfront.net/live/index.m3u8
--- a/streams/iq.m3u
+++ b/streams/iq.m3u
@@ -1,4 +1,6 @@
 #EXTM3U
+#EXTINF:-1 tvg-id="4kurdTV.iq",12 4kurd TV (1080p)
+https://4kurd.icdndhcp.com/hls/stream.m3u8
 #EXTINF:-1 tvg-id="12ArabiaTV.iq",12 Arabia TV (1080p) [Geo-blocked]
 https://livestream.almanasa.tv:8029/news/ch19/ch19_1080.m3u8
 #EXTINF:-1 tvg-id="24iraqTV.iq",24 iraq TV (1080p) [Geo-blocked]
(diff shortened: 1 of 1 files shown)


streams/iq.m3u
```diff
@@ -1,4 +1,6 @@
 #EXTM3U
+#EXTINF:-1 tvg-id="4kurdTV.iq",12 4kurd TV (1080p)
```
freearhey, line 2:
```suggestion
#EXTINF:-1 tvg-id="4KurdHD.fr",12 4kurd TV (1080p)
```

freearhey (changes requested):

streams/iq.m3u
Redo-San: Please, it's an Iraqi Kurdish music channel, not a French channel. I think you're confused.

freearhey: Channels are sorted by the country specified in the [iptv-org/database](https://github.com/iptv-org/database).

Redo-San: > Channels are sorted by the country specified in the [iptv-org/database](https://github.com/iptv-org/database).

Yes I know that and I think there is an error sorting this channel.

BellezaEmporium: Nope. 4Kurd is a channel in destination of the Iraqi people indeed, but it is from a French group named Persiana. Thus, this channel is based in France, not in Iraq.

Redo-San: Most of the Iraqi channels establish companies and offices in Europe, America and some countries neighboring Iraq to avoid being closed by the authorities or stormed by protesters who oppose their approach and policies.

Therefore, when you read that an Iraqi channel is based in America, this does not mean that it is an American channel, for example Al-Hurra channel.

The same thing here in 4Kurd, it is a Kurdish channel, but because its founder is from the Iranian opposition, they are forced to broadcast from France.

For this reason, it is wrong to categorize the channel according to the country in which its office or company is located, because the channel does not necessarily speak to the country in which it is located.

BellezaEmporium: While I understand your point of view, we can not do such an exception while the rules are made like this, since this channel *isn't streamed from Iraq*, but *aimed at Iraqis*.

The structure of our playlists are made as such : all the channels that you may see on our playlists are sorted in such way that defines from where they are streamed.

While it is "forced to broadcast from France" and aimed at a different group of people and/or diaspora than expected, it remains a channel that is based and streamed in France. Thus, it is a "French channel made for Iraqis".

We have a few channels that fall under that same dilemma, and we've decided to sort them under the country in which they are based at.

Redo-San: Well let's end this discussion
Put the channel in the French playlist, and let the French enjoy the beautiful Kurdish music.
But you should know that this decision is unfair to the channel and its followers.

freearhey: Just to clarify, it doesn't matter which file you put the links in. 
The only thing that affects which playlist a given channel will be in is the `broadcast_area` specified in the [iptv-org/database](https://github.com/iptv-org/database). In the case of 4kurd TV it is Iran: https://iptv-org.github.io/channel?id=4KurdHD.fr If you think this information is incorrect you can leave a request to change it [here](https://github.com/iptv-org/database/issues/new?assignees=&labels=channels%3Aedit&projects=&template=channels_edit.yml&title=Edit%3A+) and if it is approved, the channel will automatically be moved to another playlist. 
And to make it all work we ask you to specify the correct `tvg-id`.

Redo-San: The Kurds do not have a homeland or an official state until we classify them in it. They live in Iraq, Iran, Turkey and Syria, and they speak the same language. The only official thing they have is the Iraqi Kurdistan region, in which they enjoy a self-governing system independent of the Iraqi state.

For this reason, we always put Kurdish channels in the Iraqi playlist, because Iraq has the only official, internationally recognized gathering of Kurds.

But if we were to classify the 4kurd channel according to the nationality of the person who owns it, this means that we would classify it as an Iranian channel because the owner of the channel is an Iranian named Majid Yousefy, while the channel does not speak the Persian language, but rather the Kurdish language.

As for [submitting a request to change the channel information](https://github.com/iptv-org/database/issues/new?assignees=&labels=channels%3Aedit&projects=&template=channels_edit.yml&title=Edit%3A+), believe me, I no longer care about the matter. I will delete the channel from the Iraqi playlist, and I will leave in peace.

freearhey (changes requested):
Please recheck all other `tvg-id` as well, as they don't match the ones in the database either.

Redo-San:
> Please recheck all other `tvg-id` as well, as they don't match the ones in the database either.

I am fed up with your database full of errors, mark the channel that you think has a problem and I will amend it.

freearhey:
I have no idea what database errors you are talking about. I only asked you to check the `tvg-id` you specified in your request and make them match the ones in the database.

Redo-San:
I checked it and there is no problem. I am Iraqi and I live in Iraq and I know these channels very well. But if they do not match your database, then this is not my fault.
I will close this request today because you argue every time to check the tvg-id in a provocative and inconsiderate manner.

PR #14471, opened 2023-10-17
Remove AlJazeeraChannel.qa from 480p link so https://iptv-org.github.io/iptv/languages/ara.m3u contains the 1080p link

CasperMcFadden95:


BellezaEmporium (commented):
It gets the higher quality stream first, so I do not see the point into removing this, apart of getting the EPG off one channel.

CasperMcFadden95:
> It gets the higher quality stream first, so I do not see the point into removing this, apart of getting the EPG off one channel.

Thanks but https://iptv-org.github.io/iptv/languages/ara.m3u doesn't contain the 1080p link so there might be a bug in the script. It looks like it picks a mirror at random. "LTV (540p)" is another example even though there's a 720p link in sy.m3u

PR #14474, opened 2023-10-18
Update br.m3u

allanf181:
https://github.com/iptv-org/database/pull/6731

### DIFF
--- a/streams/br.m3u
+++ b/streams/br.m3u
@@ -79,13 +79,17 @@ https://arkyvbre1g.zoeweb.tv/fiocruz/fiocruz.stream/playlist.m3u8
 http://189.51.193.70:1935/canalsete/xpl-c7/playlist.m3u8
 #EXTINF:-1 tvg-id="CanalTVRio.br",Canal TV Rio (720p)
 https://video01.logicahost.com.br/canaltvrio/canaltvrio/playlist.m3u8
-#EXTINF:-1 tvg-id="CartoonNetworkBrazil.br",Cartoon Network Brazil (720p)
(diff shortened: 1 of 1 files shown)


streams/br.m3u
```diff
@@ -121,6 +125,10 @@ http://www.portalcultura.com.br/playerhtml/funtelpa/tv_funtelpa/playlist.m3u8
 https://59d39900ebfb8.streamlock.net/cwbtv/cwbtv/playlist.m3u8
 #EXTINF:-1 tvg-id="DemaisTV.br",Demais TV (720p)
 https://stmv1.samcast.com.br/demaistv6503/demaistv6503/playlist.m3u8
+#EXTINF:-1 tvg-id="DiscoveryKidsLiveBrazil.br",Discovery Kids Brazil (720p)
```
freearhey, line 128: "Discovery Kids Brazil" is on the blocklist due to claims of copyright holders (https://github.com/iptv-org/iptv/issues/1831)

https://github.com/iptv-org/iptv/actions/runs/6556381299/job/17814840013?pr=14474

freearhey (changes requested):

streams/br.m3u
BellezaEmporium: Counter intuitive nonetheless, when the actual stream is public.

allanf181: The stream is different from that transmitted by satellites that use encryption, so this shouldn't apply. I think you could put an extra name next to the name of the stream to differentiate it, perhaps FreeLive, which is what is written in the URL of the site.

freearhey (approved):

PopeyeTheSai10r (approved):

PR #14518, opened 2023-10-24
Decouple Ucom source

Dum4G:
Decouple of Ucom source into a separate playlist. The problem with Ucom is that most links from it were coming from 2019-2021 leaks and seems like nobody tried to actualise it since then. Some channels are now gone, some got renamed and some listings are now showing absolutely different channels. As of now we can do the part one and perform a simple decoupling, but it would better to also find an automated way to check if a link still working and possibly to get some metadata about the channel


Dum4G:
image
In a situation like that shall the channel playlist be prevailed over source playlist? @freearhey

Dum4G:
I'm done for now then. Ready for review

freearhey (dismissed):

freearhey (approved):

BellezaEmporium (approved):

PR #14922, opened 2023-11-01
Patch 2023.11.1

freearhey:
Changes:

- added support for `tvg-shift` attribute
  
  If the broadcast is delayed compared to normal broadcasting, this can now be indicated in the description as follows:
  ```
  #EXTINF:-1 tvg-id="ExampleTV.au" tvg-shift="6",Example TV +6
  http://example.com/index.m3u
  ```
- updated algorithm for sorting links in public playlists
  
  Now all links are sorted by the following attributes:
  - channel ID (ascending)
  - timeshift (ascending)
  - quality (descending)
  - label (ascending)

  This should fix the situation where low quality broadcasts are prioritized over high quality (https://github.com/orgs/iptv-org/discussions/1522).

- ~~removed local channels from the country playlists~~
   
  ~~Channels with only state/province specified in the broadcast area in [iptv-org/database](https://github.com/iptv-org/database) will be included only in the playlist of this subdivision.~~


Carlinhos027:
I think the last point is a mistake. I think it's better that they are all on the same list, otherwise it would be a mess.

CasperMcFadden95:
Is it possible to generate a full playlist for a certain language with all duplicates? Thanks

Dum4G:
I'm glad to see feedback being addressed. But I would like to see a few more changes if it's still debatable and possible:

1. To automatically assign timeshifted channels into Timeshifted category and sort them in lesser priority than the rest of other channels. In my particular case I'm able to add at least 91 timeshifted sources and I'm not doing this to avoid growing clutter of the playlist
2. Regional channels may also be present in per-country playlists because most people are likely sticking to a single playlist and doesn't use any advanced players that can merge every playlist of users choice. Proposed solution is the same here: to separate them into a special category with the least priority in arrangement

freearhey:
@Carlinhos027 Just to clarify, for example we have a link:

https://github.com/iptv-org/iptv/blob/2aee478eb8b3bfc6c31ddaa957d009b31553cd9a/streams/us_local.m3u#L2-L3

Right now when generating public playlists it goes at least into two playlists:

United States: https://iptv-org.github.io/iptv/countries/us.m3u

and

California: https://iptv-org.github.io/iptv/subdivisions/us-ca.m3u

In the new version, it will only go on the California playlist, and the US playlist will only include channels intended to be broadcast nationwide.

Do you think it's worse this way?

freearhey:
@CasperMcFadden95 I'm sorry I didn't understand the question. Could you give me an example?

freearhey:
@Dum4G 

> In my particular case I'm able to add at least 91 timeshifted sources

Could you be more specific about which sources you are referring to?

> most people are likely sticking to a single playlist and do not use any advanced players that can merge every playlist of users choice

Good point!

Dum4G:
> Could you be more specific about which sources you are referring to?

Mainly streams from Zabava (Rostelecom) for some popular channels that broadcast for several timezones and official ones from NTV, more to discover?

freearhey:
Сhanges related to local channels in the country playlist are removed from PR.

freearhey:
> Mainly streams from Zabava (Rostelecom) for some popular channels that broadcast for several timezones and official ones from NTV, more to discover?

Well for now you can just add the `tvg-shift` attribute to all such links, and if it does become a problem we can use this attribute to easily separate them from the rest of the links.

Dum4G:
image
image

If I were insane enough to use this find even reverted commits might have become handy.

CasperMcFadden95:
> @CasperMcFadden95 I'm sorry I didn't understand the question. Could you give me an example?

Thanks for respond.

https://iptv-org.github.io/iptv/languages/ara.m3u only contains one `MBC1.ae` link.
```
#EXTINF:-1 tvg-id="MBC1.ae" tvg-logo="https://i.imgur.com/ePBvFC8.png" group-title="General",MBC 1 (1080p)
https://d3o3cim6uzorb4.cloudfront.net/out/v1/0965e4d7deae49179172426cbfb3bc5e/index.m3u8
```

 Is it possible to generate another ara.m3u that contains multiple mirrors for the same tvg-id.

```
#EXTINF:-1 tvg-id="MBC1.ae",MBC 1 (1080p)
https://d3o3cim6uzorb4.cloudfront.net/out/v1/0965e4d7deae49179172426cbfb3bc5e/index.m3u8
#EXTINF:-1 tvg-id="MBC1.ae",MBC 1 (1080p)
https://mbc1-enc.edgenextcdn.net/out/v1/0965e4d7deae49179172426cbfb3bc5e/index.m3u8
#EXTINF:-1 tvg-id="MBC1.ae",MBC 1 (1080p)
https://shls-live-ak.akamaized.net/out/v1/0965e4d7deae49179172426cbfb3bc5e/index.m3u8
```

Thanks

scripts/commands/playlist/generate.ts
```diff
@@ -104,7 +104,15 @@ async function loadStreams({
   let streams = await parser.parse(files)
 
   streams = streams
-    .orderBy([(stream: Stream) => stream.channel, (stream: Stream) => stream.url], ['asc', 'asc'])
+    .orderBy(
+      [
+        (stream: Stream) => stream.channel,
+        (stream: Stream) => stream.timeshift,
+        (stream: Stream) => parseInt(stream.quality.replace('p', '')),
+        (stream: Stream) => stream.label
```
CasperMcFadden95, line 112: Maybe you can sort by host as well? So that a YouTube link is only used if there's no other option. Thanks

freearhey: What's wrong with the YouTube link?

CasperMcFadden95: It requires an extractor like https://ythlsgo.onrender.com/channel/ which goes down

freearhey: Well if the link doesn't work it should just be removed from the playlist

CasperMcFadden95: Wouldn't it reduce the load on the extractor and thus it won't go down as much?

freearhey: I don't know, it's better to ask the author of this service about it.

freearhey: Broken links removed in https://github.com/iptv-org/iptv/pull/14940

BellezaEmporium (approved):
Now that I see everyone's alright with the PR, approving.

Alstruit (approved):

PR #15019, opened 2023-11-12
Create us_iptvmate.m3u

freearhey:


Dum4G (approved):

BellezaEmporium (changes requested):
Quite reticent in the idea of adding this list, as those are merely redirectors to RedTraffic's feeds of the same name.

(btw, if we can consider those "feeds", they all redirect to "mycamtv.com" feeds.)

PR #15149, opened 2023-11-25
Add alternative YouTube streams for Alhurra and Alhurra Iraq

AppleSheeple:
Official website stream (already in the playlist) has issues
 (jittery, breaks out).

 No such issues from the official YouTube streams. So add those as
 alternatives.

 Didn't find info about alternative streams regarding naming, so
 just added "(Alternative)" to the name.

### DIFF
--- a/streams/us_ythls.m3u
+++ b/streams/us_ythls.m3u
@@ -121,6 +121,10 @@ https://ythls.armelin.one/channel/UCROKYPep-UuODNwyipe6JMw.m3u8
 https://ythls.armelin.one/channel/UCbKWv2x9t6u8yZoB3KcPtnw.m3u8
 #EXTINF:-1 tvg-id="AlHiwarTV.uk",Alhiwar TV (1080p) [Not 24/7]
 https://ythls.armelin.one/channel/UC_fNP2ettM5ius4BZ6G7MOg.m3u8
+#EXTINF:-1 tvg-id="AlhurraAlt.us",Alhurra (Alternative) (720p)
+https://ythls.armelin.one/channel/UCyscVWiJELkATSuU-RF2NLg.m3u8
+#EXTINF:-1 tvg-id="AlhurraIraqAlt.us",Alhurra Iraq (Alternative) (720p)
+https://ythls.armelin.one/channel/UCTkMv4UVWC7bpnO1edsrB9A.m3u8
 #EXTINF:-1 tvg-id="AlltheKPop.kr",All The K-POP (1080p) [Not 24/7]
 https://ythls.armelin.one/channel/UCPde4guD9yFBRzkxk2PatoA.m3u8
 #EXTINF:-1 tvg-id="AlmagharibiaTV.uk",Almagharibia TV (1080p)


AppleSheeple:
I think I've done it right eventually with the change from the database side.

freearhey:
If the official broadcast is bad why not just remove it from the playlist then?

AppleSheeple:
> If the official broadcast is bad why not just remove it from the playlist then?

Because they may fix it... eventually!

If it is preferred to keep only one stream per channel, then I will remove the official ones.

streams/us.m3u
```diff
@@ -109,8 +109,12 @@ https://stream.al-mahdi.tv/hls/1234.m3u8
 http://live.albavision.net:1123/live/albdreams.m3u8
 #EXTINF:-1 tvg-id="Alhurra.us",Alhurra (720p)
 https://mbn-ingest-worldsafe.akamaized.net/hls/live/2038900/MBN_Alhurra_Worldsafe_HLS/master.m3u8
+#EXTINF:-1 tvg-id="AlhurraAlt.us",Alhurra (Alternative) (720p)
+https://ythls.armelin.one/channel/UCyscVWiJELkATSuU-RF2NLg.m3u8
```
CasperMcFadden95, line 113: Not working

streams/us.m3u
```diff
@@ -109,8 +109,12 @@ https://stream.al-mahdi.tv/hls/1234.m3u8
...
 #EXTINF:-1 tvg-id="AlhurraIraq.us",Alhurra Iraq (720p)
 https://mbn-ingest-worldsafe.akamaized.net/hls/live/2038899/MBN_Iraq_Worldsafe_HLS/master.m3u8
+#EXTINF:-1 tvg-id="AlhurraIraqAlt.us",Alhurra Iraq (Alternative) (720p)
+https://ythls.armelin.one/channel/UCTkMv4UVWC7bpnO1edsrB9A.m3u8
```
CasperMcFadden95, line 117: Not working

streams/us.m3u
AppleSheeple: Sort of. It's the redirector's fault. I didn't notice because it's working for me connecting from Europe. But someone in the middle east told me it doesn't work reliably for them.

I'd imagine this would be an issue with any https://ythls.armelin.one link.

And the official stream is not fine. Watch for more than a few seconds and you will notice.

streams/us.m3u
```diff
@@ -109,8 +109,12 @@ https://stream.al-mahdi.tv/hls/1234.m3u8
 http://live.albavision.net:1123/live/albdreams.m3u8
 #EXTINF:-1 tvg-id="Alhurra.us",Alhurra (720p)
 https://mbn-ingest-worldsafe.akamaized.net/hls/live/2038900/MBN_Alhurra_Worldsafe_HLS/master.m3u8
+#EXTINF:-1 tvg-id="AlhurraAlt.us",Alhurra (Alternative) (720p)
```
freearhey, line 112:
```suggestion
#EXTINF:-1 tvg-id="Alhurra.us",Alhurra (720p)
```

streams/us.m3u
```diff
@@ -109,8 +109,12 @@ https://stream.al-mahdi.tv/hls/1234.m3u8
...
+https://ythls.armelin.one/channel/UCyscVWiJELkATSuU-RF2NLg.m3u8
 #EXTINF:-1 tvg-id="AlhurraIraq.us",Alhurra Iraq (720p)
 https://mbn-ingest-worldsafe.akamaized.net/hls/live/2038899/MBN_Iraq_Worldsafe_HLS/master.m3u8
+#EXTINF:-1 tvg-id="AlhurraIraqAlt.us",Alhurra Iraq (Alternative) (720p)
```
freearhey, line 116:
```suggestion
#EXTINF:-1 tvg-id="AlhurraIraq.us",Alhurra Iraq (720p)
```

freearhey (changes requested):
Also, we store links from ythls in a separate file:: https://github.com/iptv-org/iptv/blob/master/streams/us_ythls.m3u

streams/us.m3u
AppleSheeple: Moved to `us_ythls.us`.

But this suggestion wouldn't work because it's the same `tvg-id` as the official stream!

freearhey (approved):

Dum4G (approved):

streams/us.m3u
Dum4G: seems fine now

AppleSheeple:
The accompanying database PR is still open!

*  iptv-org/database#8105

PR #15261, opened 2023-12-09
[TEST] Fix workflows

BellezaEmporium:
Node doesn't consider .ts files like .js files, as default. Since we aren't working in an ES6 module, try and fix the issue by calling ts-node-esm. This *doesn't change any dependencies* by far, only loads the file from another method.

EDIT : Seems that v18.19.0 implemented certain changes from v20, or some similar changes. Issue here is common and already discussed in ts-node's issues.


BellezaEmporium:
#15262

PR #15451, opened 2023-12-29
Add Top News Albania

julianxhokaxhiu:
It uses a proxy url to obtain the official Twitch streaming m3u8 playlist.

### DIFF
--- a/streams/al.m3u
+++ b/streams/al.m3u
@@ -29,7 +29,7 @@ http://198.244.188.94/panorama/livestream/playlist.m3u8
 https://deb10stream.duckdns.org/hls/stream.m3u8
 #EXTINF:-1 tvg-id="SitiShqipTV.al",Siti Shqip TV (720p) [Not 24/7]
 https://vs.sednastream.com:1936/sitishqip/sitishqip/playlist.m3u8
-#EXTINF:-1 tvg-id="Top-Channel.tv",Top News (720p)
+#EXTINF:-1 tvg-id="TopChannel.al",Top News (720p)
 https://trueodin.serv00.net/?t=topnewsal
 #EXTINF:-1 tvg-id="TropojaTelevizion.al",Tropoja TV (1080p)
 https://live.prostream.al/al/smil:tropojatv.smil/playlist.m3u8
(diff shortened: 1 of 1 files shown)


BellezaEmporium (approved):

streams/al.m3u
```diff
@@ -29,6 +29,8 @@ http://198.244.188.94/panorama/livestream/playlist.m3u8
 https://deb10stream.duckdns.org/hls/stream.m3u8
 #EXTINF:-1 tvg-id="SitiShqipTV.al",Siti Shqip TV (720p) [Not 24/7]
 https://vs.sednastream.com:1936/sitishqip/sitishqip/playlist.m3u8
+#EXTINF:-1 tvg-id="Top-Channel.tv",Top News (720p)
```
BellezaEmporium, line 32:
```suggestion
#EXTINF:-1 tvg-id="TopChannel.al",Top News (720p)
```

BellezaEmporium (dismissed):
See above.

Dum4G:
Can't confirm it to be a great implementation. Returns 5 links that either fail to load after "commercial break" or load sound only in VLC. SimpleTV failed to open it at all. So may not be a good choice on many devices and players

julianxhokaxhiu:
> Can't confirm it to be a great implementation. Returns 5 links that either fail to load after "commercial break" or load sound only in VLC. SimpleTV failed to open it at all. So may not be a good choice on many devices and players

Can you please try it on another stream? I did try it on various ones and it works just fine, but the topnewsal channel atm is having issues even opening it on Twitch. Returning a list of streams should be fine as they are are ordered from best to worst so ideally if the first doesn't work, it fallbacks to the second and so on. Also the m3u8 returned is up to the standard so I think any player should be able to handle that.

Dum4G (approved):

BellezaEmporium (commented):
2 things on my side : 

-> No adaptive playlist is returned by the Twitch website it seems. Is it a old variant of query-streamlink in the backend ?
-> My antivirus flags the website as "suspicious activities detected". Rather odd.

BellezaEmporium (approved):

BellezaEmporium:
My implementation gives a Twitch bumper saying "Preparing your stream... Watch a stream in it's natural habitat" before going in the stream.

BellezaEmporium:
https://q-s.bellezaemporium.serv00.net/iptv-query?url=https://www.twitch.tv/topnewsal

if you need a check.

Dum4G:
Same implementation, same result

julianxhokaxhiu:
Thank you a lot :)

BellezaEmporium:
> Same implementation, same result

Seems like it's a Twitch-only issue.

PR #15454, opened 2023-12-30
Updated URLs to decode brackets.

xemles:


Dum4G:
Does it help to fix anything? There was a commit that does an exact the opposite change to make some links work on some devices ig
https://github.com/iptv-org/iptv/pull/15426

xemles:
On my end it seemed to cause an issue where links would show up as invalid.

And all the other links already have brackets and aren't encoded as such.

I didn't see the one from a few days ago though. It's interesting how not all the links seemed to cause an issue for them.

Dum4G:
@freearhey what would we consider?

freearhey:
There is no need to encode links in playlists.

Dum4G (approved):

PopeyeTheSai10r:
> And all the ones with proper brackets worked well in other lists.

Please share more information about your setup. I would like to try to replace this if possible
 Which IPTV Player are you using?

BellezaEmporium (approved):

PR #15463, opened 2023-12-31
Clean Stirr & Delete Bumblebee

brokenstream:
Inspect element is showing that many of these stream's sources lead to a 404 page. None of which have any replacements to be found in the Stirr API.

Meanwhile, Bumblebee/ViewNexaTV no longer have any live channels.


Dum4G (approved):

BellezaEmporium:
> Inspect element is showing that many of these stream's sources lead to a 404 page. None of which have any replacements to be found in the Stirr API.
> 
> Meanwhile, Bumblebee/ViewNexaTV no longer have any live channels.

Needs a verification onto Frequency.com / Frequency.stream, since the Google DAI links are only redirectors.

BellezaEmporium (approved):

PR #15624, opened 2024-01-21
Update au.m3u

AdamXweb:
Update 9 Gem stream to gem instead of current channel 9 stream

### DIFF
--- a/streams/au.m3u
+++ b/streams/au.m3u
@@ -8,7 +8,7 @@ https://6n3yogbnd9ok-hls-live.5centscdn.com/threetamil/d0dbe915091d400bd8ee7f27f
 #EXTINF:-1 tvg-id="6PR.au",6PR Perth (1080p)
 https://6prlive.akamaized.net/hls/live/2033806/6PR/index.m3u8
 #EXTINF:-1 tvg-id="9GemSydney.au",9Gem (720p) [Geo-blocked]
-https://9now-livestreams.akamaized.net/hls/live/2007330/ch9-syd/master.m3u8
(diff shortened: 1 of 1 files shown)


Alstruit (dismissed):

streams/au.m3u
```diff
@@ -8,7 +8,7 @@ https://6n3yogbnd9ok-hls-live.5centscdn.com/threetamil/d0dbe915091d400bd8ee7f27f
 #EXTINF:-1 tvg-id="6PR.au",6PR Perth (1080p)
 https://6prlive.akamaized.net/hls/live/2033806/6PR/index.m3u8
 #EXTINF:-1 tvg-id="9GemSydney.au",9Gem (720p) [Geo-blocked]
-https://9now-livestreams.akamaized.net/hls/live/2007330/ch9-syd/master.m3u8
```
Dum4G, line 11: If it's still a working link a separate channel should be created

AdamXweb: The stream I'm replacing is a duplicate of "[Channel 9](https://github.com/iptv-org/iptv/blob/ec422f1070309c40c39366fada9ab7dd1155f65f/streams/au.m3u#L52)" 

The streaming content for 9 Gem should be the stream I've linked, not what was there. 

The difference between 9 Sydney and Melbourne would only be ads and news and I don't think it is necessary have its own stream, if so that would be separate to this change.

Dum4G: It's fine to fix Sydney and Melbourne inconsistencies in this PR

Dum4G (approved):

Alstruit (approved):

PR #15863, opened 2024-02-13
Update am_ucom.m3u

AntiPontifex:
See title


Udihg (commented):
Ff

Dum4G:
I do like the tvg-id part but what's wrong with local naming?

AntiPontifex:
Nothing is wrong with the local naming: I was attempting to establish a standard with the Roman alphabet. We do have the Alternate Name field in the database to accommodate for other alphabets. Would it be amiss to make the Roman alphabet standard in other .m3us as well? Please discuss.

AntiPontifex:
Discussion?

Dum4G:
I believe local namings is what language-speaking viewers expect and what's fine for now. My other thoughts go way over current implementation.

AntiPontifex:
Unfortunately, non-Roman alphabets do not meet ISO standards: therein lies the problem.

Dum4G:
> Unfortunately, non-Roman alphabets do not meet ISO standards: therein lies the problem.

For the time being, as we can't distinguish what viewers have in generated playlists and what maintainers see to help them maintain better, I would prefer to stick with naming that is native and comfortable to viewers

AntiPontifex:
My vote is maintaining the standard that IPTV github is supposedly following: ISO.

PR #15907, opened 2024-02-16
add 3 more mexican channels

ghost:


streams/mx.m3u
```diff
@@ -189,12 +191,14 @@ https://stream.oursnetworktv.com/radio/luperronaocotlan/playlist.m3u8
...
 #EXTINF:-1 tvg-id="LaVozGruperaTV.mx",La Voz Grupera TV (720p)
 https://cloudvideo.servers10.com:8081/8128/index.m3u8
+#EXTINF:-1 tvg-id="LasEstrellas.mx",Las Estrellas (576p)
+https://linear-413.frequency.stream/mt/studio/413/hls/master/playlist.m3u8
 #EXTINF:-1 tvg-id="LasEstrellasPlus1.mx",Las Estrellas +1 (1080p)
 https://linear-414.frequency.stream/mt/studio/414/hls/master/playlist.m3u8
 #EXTINF:-1 tvg-id="LasEstrellasPlus2.mx",Las Estrellas +2 (1080p)
 https://linear-415.frequency.stream/mt/studio/415/hls/master/playlist.m3u8
-#EXTINF:-1 tvg-id="LasEstrellasLatinAmerica.mx",Las Estrellas Latin America (720p)
-https://channel01.akamaized.net/hls/live/2022749/event01/index.m3u8
```
Dum4G, line 197: it works

Dum4G (changes requested):

streams/mx.m3u
c4artisan: but it's not las estrellas latin america (which is separate from the mexican/original). i just noticed it's a weird stream mixing up both the original feed and the latam feed for when the original is airing infomercials. i'll readd it as the og las estrellas. also, why did you review only the removal, when i added an actual stream of the feed after?

c4artisan: but it isn't las estrellas latin america, it was a stream of xew (mexican las estrellas main station) which is why i put the stream as las estrellas mexico. the original stream is a mix of both mexico and latam? for when las estrellas mexico is airing infomercials. when i saw that it was like that i got confused and put up a different stream in las estrellas' place on the list but now i'm going to revert that (changing las estrellas mexico) because it works, just like you said

c4artisan: i just did it, so i suggest you take a look at the comment above and see that you requested changes without knowing the stream wasn't of the channel it was said to be, and maybe dismiss your review

Dum4G (approved):

Alstruit (approved):

PR #15984, opened 2024-02-27
Update pt.m3u

pimboli:
added Euronews PT


Dum4G:
None of those links open in VLC or browser for me. And even if there's a chance to watch them we have https://github.com/iptv-org/iptv/blob/master/streams/fr_euronews.m3u

PR #15994, opened 2024-02-27
[PROPOSAL] Few clarifications

BellezaEmporium:
- Rewrite the "No NSFW" sentence.
- I believe we should be **way** stricter concerning paid channels. If the channel is seen as a PPV or paid channel, it **won't** be allowed, in order to keep the playlists clean from obvious copyright claims.
- Clarify our position about Xtream.


guaymatetv:
https://glb.bozztv.com/glb/ssh101/guaymatetv/index.m3u8

PR #16208, opened 2024-03-19
Update iq.m3u - Zarok TV Sorani added (completely different channel t…

ozgurkazancci:
…han Turksat's Zarok TV)

Zarok TV Sorani added (completely different channel than Turksat's Zarok TV)


BellezaEmporium (approved):

BellezaEmporium:
While ZarokTV.iq doesn't exist in the database, I still approve it because there's indeed a difference in those two. If you can add this channel in iptv-org/database that would be perfect.

Dum4G (approved):

ozgurkazancci:
> While ZarokTV.iq doesn't exist in the database, I still approve it because there's indeed a difference in those two. If you can add this channel in iptv-org/database (via an issue) that would be perfect.

Done!

PR #16529, opened 2024-05-10
Remove autogenerated files

Alstruit:
Should resolve conflicts when running npm commands on initilization.


BellezaEmporium:
The .lock file is to make sure it freezes the dependencies of the project at last working state (when pushed).

Where's the issue in keeping this file ?

monyasau:
since this is an unneeded pull request can't it kindly be deleted @Alstruit

Issue #16689, opened 2024-06-04
Broken: Cartoon Network Russia


iamdelickkk:
broken url  - 404 error
new url - `http://stream01.vnet.am/Kinosemya/mono.m3u8`

azgaresncf:
**Copyrighted stream.**

azgaresncf:
nvm, i didn't see the vnet.m3u8 file :')

noreplywavetv-eng:
Not working anymore 😭

PR #16701, opened 2024-06-05
Fix Cartoon Network Russia stream link

ghost:
Apparently, the stream didn't load, so i had to replace with new one
Issue: #16689
Tested with mpv and streams loads fine


PopeyeTheSai10r:
Both links are working for me.
Instead of replacing the old one. I will recommend we add a new line with the new one.

indeededa:
i forgot to close that pull request, because these links are working, maybe i would add same channel with diffierent stream link one

PR #16950, opened 2024-07-04
Update ve.m3u

zzainss:
agregado tam merida


BellezaEmporium (changes requested):
Incorrect tvg-id.

Please take a look at https://iptv-org.github.io for the corresponding tvg-id. If it doesn't exist, you can also create a [database entry](https://github.com/iptv-org/database) in order to add it, or keep the tvg-id blank.

Dum4G:
image

PR #17019, opened 2024-07-16
Update channels every hour

ghost:
Why wait sometimes up to a day when you can update channels every hour?


Dum4G:
We don't make changes that often. Also, an overwhelming usage of Github Actions once almost caused permanent ban of this repository

GuyInCanada:
If you want channels updated every hour, fork the repo and maintain it yourself.

PR #17060, opened 2024-07-24
Update ca_pluto.m3u

iotv0815:
fill the empty tvg-id with the channel name


Sphinxroot (commented):
please read https://github.com/iptv-org/database/blob/master/CONTRIBUTING.md  and tvg-id must fit in https://github.com/iptv-org/database

PR #17068, opened 2024-07-24
Update cn.m3u

iotv0815:
fill empty tvg-id with the channel's name so to make it searchable in client


Sphinxroot (commented):
please read https://github.com/iptv-org/database/blob/master/CONTRIBUTING.md  and tvg-id must fit in https://github.com/iptv-org/database

PR #17088, opened 2024-07-27
Update ru.m3u

smolnp:
Russian Russian channel groups have been translated, inactive channels have been removed, new channels have been added, and Kazakh channels that are broadcast in Kazakh without using Russian have been removed.


Dum4G:
Please read https://github.com/iptv-org/iptv/blob/master/CONTRIBUTING.md, specifically Project Structure. The playlist you're trying to edit should not contain channel info other than tvg-id and stream, it's only used for aggregation. If you know Kazakh channels that do not have Russian subbing but still are in the playlist please let me know which one

PR #17234, opened 2024-08-12
Create us_moveonjoy.m3u

KevinPfeiffer:
BTW HBO 2 is excluded due to the new policy not allowing Warner Bros. Discovery channels (including HBO-branded channels) to be part of iptv-org anymore


Alstruit:
I have merged the changes onto #17235. Please make further changes from there.

PR #17237, opened 2024-08-12
Update iq.m3u

Hani-Poo:
Adding Karbalatv from iraq Updating iq.m3u


Dum4G:
Please move into https://github.com/iptv-org/iptv/blob/master/streams/us_ythls.m3u

Issue #17582, opened 2024-09-21
Broken: NHL Network


Rinbusts:
### Broken Link

https://lnc-nhl.tubi.video/playlist.m3u8

### What happened to the stream?

Not loading

### Notes (optional)

Crashes player when trying to open, no other nhl channels appear when they should.

### Contributing Guide

- [X] I have read [Contributing Guide](https://github.com/iptv-org/iptv/blob/master/CONTRIBUTING.md)

Goldiefresh:
Here's the same channel from a different source. 

https://nhl-firetv.amagi.tv/playlist.m3u8

R4IN2007:
https://lnc-nhl.tubi.video/playlist.m3u8 - link offline

vtjapes:
> Here's the same channel from a different source.
> 
> https://nhl-firetv.amagi.tv/playlist.m3u8

This is not NHL Network by the way.  It's just an NHL streaming channel playing the same videos over and over.

PR #17626, opened 2024-09-28
Add NHL Channel [Amagi Source]

ghost:
NHL Streams in Tubi source got 404'd, so i grabbed from Amagi source and it plays fine
Until someone fixed an Tubi stream link
#17582


luuctuyen:
Wait, i forgot to close this PR because NHL Channel is already added in us_firetv.m3u with the same link

PR #17739, opened 2024-10-14
Add 2 Vietnamese VTV channels

ghost:
Stream links are grabbed from CVmedia and plays fine with MPV


Dum4G:
Plays for one second and disconnects. Need confirmation if it's the VLC issue or protection of a kind

luutuyen2k9:
It just possibly an VLC issue
Maybe can we play streams with mpv?

Dum4G:
Previously I have raised a discussion on the topic
https://github.com/orgs/iptv-org/discussions/1632

Currently I'm rejecting links that do not work in VLC because we were historically using it for stream checks. many IPTV players use it in it's core, it is platform agnostic etc. Anyway, this topic is still debatable

luutuyen2k9:
I'm closing this pull request here and I don't want to contribute to iptv-org/iptv here

PR #17834, opened 2024-10-22
Update do.m3u 

Pecaquito:


Sphinxroot:
i dont think  it go here 
https://gia.tv/watchssh101.php?pptv=208343

gia tv its usa based tv

and tom and jerry dont work


it might be better if it done in a unique playlist  ...  like us_giatv.m3u

Manchmal1 (commented):
Finde es klasse,danke

PR #18083, opened 2024-12-18
Update es.m3u

darknebular:
Added new channels A3 from Spain than left


freearhey (changes requested):
- invalid `tvg-id`
- unnecessary attributes: `tvg-logo`, `group-title`, `tvg-name`

PR #18613, opened 2025-01-19
Update pe.m3u

minato696:
Update Exitosa Tv


sguinetti (approved):

minato696:
The channel is Exitosa TV, so it must correspond to the correct link associated with that channel. It should not appear as Exitosa TV with a link to the RT en Español signal. This needs to be corrected and updated.

CasperMcFadden95 (approved):

PR #18635, opened 2025-01-20
Fix LibyaMostakbal.ly

CasperMcFadden95:


CasperMcFadden95:
Replaced with
https://github.com/iptv-org/iptv/issues/18639
https://github.com/iptv-org/iptv/issues/18640

PR #18857, opened 2025-01-27
Update hu.m3u

PaulGabi15:
2025


BellezaEmporium (commented):
Seems to be an Xtream-Codes server (auth method), therefore we do not allow it.

PR #19592, opened 2025-02-23
Update az.m3u

emilsavalan:
Azad TV url correction


BellezaEmporium (commented):
I do not understand the point of this change.

The adaptive links both works fine and provides streams for every type of bandwidth
image

emilsavalan:
The adaptive links inside m3u8 file do work. But neither VLC nor my TV could read this URL content, so I get an error.

CasperMcFadden95:
Current link works in VLC

PR #21013, opened 2025-04-04
Update Finnish channels

TVKaista:


lumiini:
The new links both display the following error that "This program is only available at the above address" 
error

PR #21330, opened 2025-04-23
Update jp_primehome.m3u

TvJapan:
修正


BellezaEmporium (commented):
Nothing seems to load.

Nobuchans:
> Nothing seems to load.

It doesn't seem to be cured.
When I open it, the screen turns black.

InterLance90:
nothing yet, when i opened VLC with the new playlist, it didn't work, just yet.

kdhh357:
It's useless if you can't open it. Are you sure this is a usable channel?

InterLance90:
Since I'm not a techie guy, I'm not sure.
But it's definitely a useless channel until we got new details later.

MapGuy11 (commented):
Can confirm nothing loads.

InterLance90:
This confirms this, what should we do now? We can wait for new development in this situation.

PR #21350, opened 2025-04-25
a massive update to polish channels

Jxpl9:
all canal plus channels deleted
added more polish channels


BellezaEmporium (changes requested):
I don't remember about having Xtream channels inside pl.m3u. We actually do not allow them up.

Jxpl9:
What is a good method to check if this link is xstream format

BellezaEmporium:
Usually, you'll have "hostname(:25641)/live/username/password/randomnumber.ts"

PR #21860, opened 2025-05-20
Update ua.m3u

xtvmediaukr:
Add new TV-channel "ПЕРШИЙ. XTV Media" from Ukraine

Requires: https://github.com/iptv-org/database/issues/18189


freearhey (changes requested):
https://github.com/iptv-org/iptv/pull/21860#issuecomment-2900495627

Issue #16839, opened 2024-06-21
DMCA takedown notices from Warner Bros. Discovery Inc.


freearhey:
> Hi iptv-org,
> 
> I'm contacting you on behalf of GitHub because we've received a DMCA takedown notice regarding the following content:
> 
> https://iptv-org.github.io/
> 
> We're giving you 1 business day to make the changes identified in the following notice:
> 
> https://enterprise.githubsupport.com/attachments/token/oBlvU4LNaVfyN6FCJHmI25kkB/?name=2024-06-20-warner-bros.rtf
> 
> If you need to remove specific content from your repository, simply making the repository private or deleting it via a commit won't resolve the alleged infringement. Instead, you must follow these instructions to remove the content from your repository's history, even if you don't think it's sensitive:
> 
> https://docs.github.com/articles/remove-sensitive-data
> 
> Once you've made changes, please reply to this message and let us know. If you don't tell us that you've made changes within the next 1 business day, we'll need to disable the entire repository according to our GitHub DMCA Takedown Policy:
> 
> https://docs.github.com/articles/dmca-takedown-policy/
> 
> If you believe your content on GitHub was mistakenly disabled by a DMCA takedown request, you have the right to contest the takedown by submitting a counter notice, as described in our DMCA Takedown Policy.
> 
> PLEASE NOTE: It is important that you reply to this message within 1 business day to tell us whether you've made changes. If you do not, the repository will be disabled.

olstern01:
Holy smokes that's quite a lot of channels.

olstern01:
TBS? TBS News? TBS News Dig? I'm quite sure they refer to the American TBS. The TBS in the document are Japanese channels, and WBD don't own that.

Alstruit:
Invalid channels (Not owned by Warner Bros. Discovery Inc.):

https://en.wikipedia.org/wiki/TBS_Television_(Japan)
TBS (Tokyo Broadcasting System Television, Inc)
TBS News (544p)
TBS News Dig (720p) [Not 24/7]

https://en.wikipedia.org/wiki/Traffic_Broadcasting_System
TBS eFM (1080p) [Not 24/7]
TBS Seoul (720p)

https://en.wikipedia.org/w/index.php?title=TvN_%28South_Korean_TV_channel%29
https://en.wikipedia.org/w/index.php?title=TvN_%28Asian_TV_channel%29
TVN (1080p) [Geo-blocked]
TVN (1080p)
[...]

PR #24275, opened 2025-06-19
Remove two remaining CNN channels

ghost:
Two of these channels are owned by the CNN Intl, likely to be part of the Warner Bros. Discovery

The DMCA takedown can be seen in #16839 

Channels were to be blocked in database: https://github.com/iptv-org/database/issues/18422, https://github.com/iptv-org/database/issues/18421


Alstruit (approved):

freearhey (changes requested):
I think we should wait for the channel blocking request to be considered before removing links to them from the playlist.

freearhey (approved):

PR #26211, opened 2025-08-24
Update nz.m3u

Moodie007:
Updated SkySports streams that didnt work - error 500


Alstruit:
We don't allow xtream codes. See [Why don't you accept links to Xtream Codes server?](https://github.com/iptv-org/iptv/blob/master/FAQ.md#why-dont-you-accept-links-to-xtream-codes-server)

PR #26649, opened 2025-09-04
Update workers

MHSanaei:
Since my Cloudflare workers hit the 100k request limit, I created a new CF account. Now I use 3 free accounts, each restricted to specific websites, so this method should last longer.


.github/workflows/check.yml
```diff
@@ -2,7 +2,7 @@ name: check
 on:
   workflow_dispatch:
   pull_request:
-    types: [opened, synchronize, reopened, edited]
+    types: [opened, synchronize, reopened]
```
Dum4G, line 5: Are you sure this file needs to be staged in this PR?

Dum4G (changes requested):

.github/workflows/check.yml
MHSanaei: Because edited makes the workflow run even when only the PR title or description changes, which causes unnecessary runs.

Dum4G (approved):

freearhey (approved):

PR #26724, opened 2025-09-07
Remove `@SD` to match tvg-id in epg

CasperMcFadden95:


freearhey:
Maybe it would be better to add `@SD` to the epg?

CasperMcFadden95:
Hi, because most of these channels are HD and not SD. Also the EPG doesn't have a seperate entry for `SD` & `HD` link in other regions.

PR #26816, opened 2025-09-13
Update id.m3u

iamutaki:
update AL-IMAN


BellezaEmporium (commented):
The issue you might have isn't an issue on my side. Removing the adaptive playlist for an direct source disables the ability for players to switch qualities, and for people with bad connections to properly watch said channel. It should normally smoothly change from one quality to another.

iamutaki:
Thanks for the clarification, that helps. I'll close the PR then. @BellezaEmporium

PR #27089, opened 2025-09-23
ID : clear out the files & fix KODIPROP

BellezaEmporium:
-> Clear out streams for ID, either 503 or 404 for linkserve, and "no host" for telkomsel.

-> KODIPROPs aren't properly done, fix that out.


BellezaEmporium:
Note that "origin" can also be filled as a KODIPROP, which means we can enable the use of certain links.

https://github.com/xbmc/inputstream.adaptive/wiki/Integration#inputstreamadaptivestream_headers

note : stream_headers can be used for m3u8 **only**, mpd files need to use inputstream.adaptive.manifest_headers.

Dum4G (approved):

Alstruit (approved):

PR #28395, opened 2025-10-21
Update lb.m3u

Mee3uk:
Updated Manar with a HD link, same source as before but they have a HD link available


CasperMcFadden95:
The player should pick the best quality based on the internet speed

PR #29867, opened 2025-12-03
Remove broken links 

Salehin-07:
Remove ~1700 broken streams based on issue #29860 and also added a directory links in streams directory that contains 2 files with 3000 broken and 8000 working links.

PR #29882, opened 2025-12-04
Tested the files again and removed all the links that are broken.

Salehin-07:
DELETED THE a new directy in /streams directory named links and as on issue #29867 . Removed all links that are broken ~1300


freearhey (changes requested):
> please ensure that CRLF is used to indicate a new line in all files.

https://github.com/iptv-org/iptv/pull/29867#pullrequestreview-3537567795

PR #31013, opened 2025-12-31
Update ve.m3u

interven:
add AS3 Sport TV


BellezaEmporium (changes requested):
Channel ID does not exist. Please add it to the database before submitting it with this tvg-id.

Archrootsda:
Added channel iptv-org/database/issues/24430
There is no necessary to change the ID

Archrootsda (approved):

BellezaEmporium (approved):

PR #31444, opened 2026-01-07
Update ma.m3u

Archrootsda:
fixed snrt links


Archrootsda:
I thank you @ipstreet312 for having pointed out this issue. at first it seemed that it did not affect traffic already redirected to the official streams cdn.live.easybroadcast.io/abr_corp/
closes in favor of  #31458

PR #32795, opened 2026-02-01
[FR] Add zén!th channel

thatsacrylic:
zén!th is airing every Sunday from 2pm to midnight in France until September 1st when it launches.


BellezaEmporium (commented):
Please add your channel to the [database](https://github.com/iptv-org/database) in order to get the tvg-id recognized.

If you keep your original tvg-id, it would consider this channel to be from Tuvalu (owning the .tv TLD & country code) to French citizens.

You can precise "Every Sunday", but not w/ the Not 24/7 pattern, as the bot wouldn't understand it.

You can write it like this : zén!th (Every Sunday) [Not 24/7]

freearhey:
https://github.com/iptv-org/iptv/blob/master/CONTRIBUTING.md#how-to-add-a-new-stream-link-to-a-playlists

PR #32824, opened 2026-02-02
Delete streams/us_moveonjoy.m3u

BellezaEmporium:
All streams are dead, so don't bother trying.

(not all, but it needs a proper checking)


Alstruit (approved):

Central302 (approved):

yelladude:
The moveonjoy streams work, that's what I'm still using for example https://fl1.moveonjoy.com/SYFY/index.m3u8

troycarpenter:
I'm at a loss here, too.  All the fl1.moveonjoy.com streams are working for me as well.

PR #33453, opened 2026-02-20
New Ghana Station added

localcodetech:
New TV stations added from ghana


Dum4G:
It would be better to create a separate file named gh_geyetv.m3u
Please add every channel into https://github.com/iptv-org/database using channel submission form. Without valid tvg-id found in the database our automatic Actions won't include those channels in GH playlist. Thank you

Issue #34383, opened 2026-03-14
Add: ColorVision.do@SD


BrayanKellyBalbuena:
### Stream ID (required)

ColorVision.do@SD

### Stream URL (required)

http://38.159.232.166:8000/play/a00z/index.m3u8

### Quality

480p

### Label

Not 24/7

### Contributing Guide

- [x] I have read [Contributing Guide](https://github.com/iptv-org/iptv/blob/master/CONTRIBUTING.md)

PR #34391, opened 2026-03-14
add ColorVision.do@SD

brayan-kelly:
Adding missing stream 

VLC Playlist
VLC playing Color Visión


Pecaquito:
https://github.com/iptv-org/iptv/issues/34383

PR #35165, opened 2026-03-28
fix: pin 5 unpinned action(s), extract 1 step output to env var

dagecko:
Re-submission of #35030. Had a problem with my fork and had to delete it, which closed the original PR. Apologies for the noise.

## Summary

This PR pins all GitHub Actions to immutable commit SHAs instead of mutable version tags and extracts a step output expression from a `run:` block into an `env:` mapping.

- Pin 5 unpinned actions to full 40-character SHAs
- Add version comments for readability
- Extract 1 step output from run block to env var (prevents shell injection via crafted filenames)

## Changes by file

| File | Changes |
|------|---------|
| format.yml | Pinned tibdex/github-app-token |
| stale.yml | Pinned tibdex/github-app-token |
| update.yml | Pinned tibdex/github-app-token, JamesIves/github-pages-deploy-action (x2) |
| check.yml | Extracted `${{ steps.files.outputs.all_changed_files }}` → `ALL_CHANGED_FILES` env var |

## How to verify

Review the diff — each change is mechanical and preserves workflow behavior:
- **SHA pinning**: `action@v3` becomes `action@abc123 # v3` — original version preserved as comment
- **Expression extraction**: `${{ steps.files.outputs.all_changed_files }}` in `run:` moves to `env:` block, referenced as `${ALL_CHANGED_FILES}` in the script — prevents shell injection if filenames contain metacharacters
- No workflow logic, triggers, or permissions are modified

I wrote a scanner called Runner Guard and open sourced it [here](https://github.com/Vigilant-LLC/runner-guard) so you can scan yourself if you want to. More on the research behind this [on Twitter](https://x.com/vigilance_one/status/2036581210663616729).

If you have any questions, reach out. I'll be monitoring comms.

\- Chris Nyhuis (dagecko)


dagecko:
Good catch on the `tr` line. The newline escape got decoded into a literal newline during my file processing. Pushed a fix — `tr '\n' ' '` is now preserved correctly on a single line, matching the original.

\- Chris (dagecko)

.github/workflows/check.yml
freearhey, line 1: It seems that the only thing that has changed here is the line-end character.

freearhey (changes requested):

dagecko:
@freearhey you're right, the diff on check.yml is mostly a CRLF to LF line ending normalization that happened while we were working on the file. Apologies for the noise in the diff.

There is one actual functional change at the bottom of the file. The step output reference was moved from inline in the run block to an env block for safer shell expansion:

Before: `npm run playlist:lint -- ${{ steps.files.outputs.all_changed_files }}`
After: Uses `env: ALL_CHANGED_FILES: ${{ steps.files.outputs.all_changed_files }}` and references it as `${ALL_CHANGED_FILES}`

This prevents the step output from being interpreted as shell code if it ever contains special characters.

If you'd prefer I resubmit with just the functional changes and without the line ending normalization, happy to do that.

- Chris

dagecko:
@freearhey cleaned up the line ending issue. The diff should now only show the actual functional changes. Thanks for flagging it.

- Chris

freearhey (changes requested):
https://github.com/iptv-org/iptv/pull/35165#discussion_r3016737838

PR #37137, opened 2026-04-25
Update br.m3u

x1colegal:
This stream is incorrect and redirects to the wrong channel.

The following link:
https://lowa8026-cmyk.github.io/Brazil/NickOnlineSpongebob.m3u8

does not play the SpongeBob channel, but instead redirects to the main NickOnline channel.

This causes incorrect playback in IPTV players, since it overrides the correct stream.

The correct stream is already present in the repository, so this PR removes the invalid duplicate.


Dum4G:
If that's the case I'd suggest just to change tvg-id to appropriate

PR #38082, opened 2026-05-22
Update M3U playlist with new channels

litoscrc:
add PlutoTV to m3u CR

### DIFF
--- a/streams/cr.m3u
+++ b/streams/cr.m3u
@@ -55,6 +55,117 @@ https://627bb251f23c7.streamlock.net:444/ExtremaKids/ExtremaKids/playlist.m3u8
 https://627bb251f23c7.streamlock.net:444/ExtremaTV/ExtremaTV/playlist.m3u8
 #EXTINF:-1 tvg-id="GAMTVcr.cr@SD",GAMTV.cr (720p)
 https://v2.azulstream.com:8081/gamtv/gamtv/index.m3u8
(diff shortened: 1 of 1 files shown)


freearhey (commented):
If these PlutoTV streams are specific to Costa Rica, I think it would be best to put them in a new file called `cr_pluto.m3u`, as it is done here: https://github.com/iptv-org/iptv/blob/master/streams/br_pluto.m3u

Archrootsda:
> If these PlutoTV streams are specific to Costa Rica, I think it would be best to put them in a new file called `cr_pluto.m3u`, as it is done here: https://github.com/iptv-org/iptv/blob/master/streams/br_pluto.m3u

@freearhey The plutotv links for the LATAM region (HISPAM) are the same for all countries in Hispanic America (Duplicate Plutotv ID). The downside is that in our repository we can not duplicate links, even when it comes to a different m3u playlist.  https://github.com/iptv-org/iptv/blob/717f0c4e38be093d377bdd2db5b735505c2d136d/streams/mx_pluto.m3u

freearhey:
If the links are already in the repository, there’s no need to add them again.

> The downside is that in our repository we can not duplicate links, even when it comes to a different m3u playlist.

If we add duplicates, it will only make maintenance more difficult, since any error will require us to go through all the files in the repository every time.

PR #38161, opened 2026-05-24
Restore M3U playlist and TV channels

alaaramadan77-hub:
Restored the M3U playlist format


Archrootsda:
tokenized links not allowed, expires

PR #38295, opened 2026-05-28
Update pt.m3u

ghost:
Disney is not free channel, some link were broken.
Added CNN too


BellezaEmporium (changes requested):
Your changes contain tokenized links.
Useless to remove links that work.
SIC and Noticias are geo-blocked, but it seems they've changed the localisation system.

PR #38652, opened 2026-06-05
Add Al Aoula Morocco

ahmadalaa0113-sudo:
AlAoula.ma@HD


ahmadalaa0113-sudo:
For @freearhey

BellezaEmporium:
Nope, 403 forbidden. You would need some autoredirection to a created tokenized link to make it work.

ahmadalaa0113-sudo:
How Can I Make It @BellezaEmporium

BellezaEmporium:
You would need to create a script that takes the SNRT pages from easybroadcast.io.

eg. https://snrt.player.easybroadcast.io/api/events/73_aloula_w1dqfwm see "stream", or "stream_no_timeshift".

Then, query for the token w/ https://token.easybroadcast.io/all?url= *the url you've found*

Once you have it, just paste it next to the URL you've found earlier in the events endpoint, and here you go.

ahmadalaa0113-sudo:
But How Can I Turn On Approve

BellezaEmporium:
That alone is not sufficient, as you're not returning a stream. The way Easybroadcast does their thing is weird.

johnshanlay14:
Please stop sending promotional or suspicious scam links to my Gmail account

PR #38667, opened 2026-06-05
jp_primehome : Delete the whole set

BellezaEmporium:
Struck again. 403 Forbidden.


Alstruit (approved):
Perhaps we should exclude this problematic website no?

BellezaEmporium:
I don't remember who's added it back. I had said it was useless to add it back because they would do it again.

Dum4G (approved):

ahmadalaa0113-sudo:
@Dum4G @BellezaEmporium Can You Guys Fix MRT 1 And MRT 2 m3u Now?

PR #38671, opened 2026-06-05
uk_bbc: Fixes

BellezaEmporium:


Dum4G (commented):
It may introduce more complexity in case if there will be an automated way to scrap the playlist or if there would be a need to quickly change the link format

BellezaEmporium:
> It may introduce more complexity in case if there will be an automated way to scrap the playlist or if there would be a need to quickly change the link format

I do not see where that complexity would be. The US playlists for example are already splitted by local & national channels and still, it's a bit messy to maintain, to the point that dead links aren't often checked out there. The HEVC links from what I see aren't the ones that are keen to change, from the looks of them, they all have the same structure.

Care to explain your thoughts a bit further ?

Dum4G:
Each time the source changes you'll have to verify two files at once or split them with additional actions if it's scrapped automatically. Whenever we separate a playlist for any source we do so to easily track links of the same syntax and most of the time of the same behavior. Also for ease of removal of entire source or to make a quick change instead of searching across multiple playlists for relevant link. I don't see how proposed change helps to resolve those purposes

BellezaEmporium:
That's understandable. So you would be into keeping the original structure we had ?

Dum4G (approved):

Alstruit (approved):

PR #39618, opened 2026-06-15
Refactor CountriesGenerator and add index summary

krataratha:
Refactored playlist generation into a helper method and added a master index summary feature.

### DIFF
--- a/scripts/generators/countriesGenerator.ts
+++ b/scripts/generators/countriesGenerator.ts
@@ -11,6 +11,12 @@ type CountriesGeneratorProps = {
   logFile: File
 }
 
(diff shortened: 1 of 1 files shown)


Dum4G:
I doubt we can call a propose to add new functions as a refactoring. [Index.m3u](https://github.com/iptv-org/iptv/blob/gh-pages/index.m3u) is already generated, so is [undefined.m3u](https://github.com/iptv-org/iptv/blob/gh-pages/countries/undefined.m3u), for jsons an [API repo exists](https://github.com/iptv-org/api/blob/gh-pages/streams.json)

PR #42421, opened 2026-06-30
Update README with Chinese translation and usage details

k2alzhang:
Translated README content to Chinese and updated usage instructions.


justavedmiddubaisky:
we do not accept this

PR #43046, opened 2026-07-04
Update us.m3u

justavedmiddubaisky:
changed this stream url to short url because there too long url


streams/us.m3u
```diff
@@ -694,7 +694,7 @@ https://fast-channels.sinclairstoryline.com/TBD/index.m3u8
...
+https://shorturl.at/Ws4iK
```
brokenstream: Would be better to use the adaptive stream instead of a redirect:
https://amg17233-amg17233c1-wisptv-us-6409.playouts.now.amagi.tv/playlist.m3u8

justavedmiddubaisky: replaced

brokenstream (approved):

Dum4G (approved):

PR #43846, opened 2026-07-08
Update ca.m3u

Sphinxroot:
channel stream change edit + add channel  and proper  referrer


freearhey (changes requested):
Merge conflict.

Sphinxroot:
something odds whit those conflict  never add this before when  adding stuff

PR #49340, opened 2026-08-22
Restore M3U playlist and add TV channels

etscooly2:
Restored the M3U playlist format and added various TV channels with their streaming URLs.


freearhey:
https://github.com/iptv-org/iptv/blob/master/CONTRIBUTING.md

PR #49370, opened 2026-08-23
Add and modifying streams (082326)

LPhilip1999:
**us_klowdtv:**
Editing assign ID for The First TV.


freearhey (approved):
Just in case, you can change the "Stream ID" using the usual form: https://github.com/iptv-org/iptv/issues/new?template=2_streams_edit.yml

brokenstream (approved):

